PR #11, opened 2017-02-14
Update the version of kubernetes' go-client

lookuptable:


lookuptable:
@wlu2016 This PR sets the correct commit sha for Kubernetes go-client. Previously we were using the git tag but bazel doesn't support that well.

controller/securenaming.go
```diff
@@ -41,15 +41,15 @@ func NewSecureNamingController(core v1core.CoreV1Interface) *SecureNamingControl
 	return &SecureNamingController{sc, si}
 }
 
-func newServiceIndexerInformer(core v1core.CoreV1Interface) (cache.Indexer, *cache.Controller) {
```
lookuptable: I'd prefer to add comments in a separate commit since it's unrelated to this commit.

I've create issue #19 to track that.

controller/securenaming.go
```diff
@@ -19,20 +19,20 @@ package controller
 import (
```
lookuptable: Tracked by #19

lookuptable:
Merging this to unblock subsequent commit. Comments have been tracked by #19 .

PR #15, opened 2017-02-16
Add cert generation lib and tool.

wlu2016:
@lookuptable

### DIFF
--- a/cert_manager/generate_cert.go
+++ b/cert_manager/generate_cert.go
@@ -44,11 +44,11 @@ type CertOptions struct {
 	// Duration that certificate is valid for.
 	ValidFor time.Duration
 
-	// Signer private key file (PEM encoded).
-	SignerPriv string
+	// Signer certificate (PEM encoded).
+	SignerCert *x509.Certificate
 
-	// Signer certificate file (PEM encoded).
-	SignerCert string
+	// Signer private key (PEM encoded).
+	SignerPriv *rsa.PrivateKey
 
 	// Juju org", "Organization for the cert.
 	Org string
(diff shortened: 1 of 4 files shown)


cert_manager/generate_cert.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type CertOptions struct {
+	// Comma-separated hostnames and IPs to generate a certificate for.
+	// This can also be set to the identity running the workload,
+	// like kubernetes service account.
+	Host string
+
+	// Creation date formatted as Jan 1 15:04:05 2011.
+	ValidFrom string
```
lookuptable: Why 'ValidFrom` is a string but `ValidFor` is a time.Duration?

cert_manager/generate_cert.go
```diff
@@ -0,0 +1,188 @@
...
+	// case, otherwise the certificate is signed by the signer private key
+	// as specified in the CertOptions.
+	priv, err := rsa.GenerateKey(rand.Reader, RsaBits)
+	if err != nil {
+		log.Fatalf("RSA key generation failed with error %s.", err)
+	}
+	template := genCertTemplate(options)
+	signingCert, signingKey := &template, priv
+	if !options.IsSelfSigned {
+		signingCert, signingKey = loadSigningCreds(options.SignerPriv, options.SignerCert)
```
lookuptable: It feels strange that the signing key and cert are passed in by the caller. Instead I think the CA should generate/read from a file during startup and use the same key & cert to sign all requests.

cert_manager/generate_cert.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Whether this cerificate is self-signed.
+	IsSelfSigned bool
+
+	// Whether this certificate is for a client.
+	IsClient bool
+}
+
+// Size of RSA key to generate.
+const RsaBits = 2048
```
lookuptable: Should this just be private?

cert_manager/generate_cert.go
wlu2016: ValidFrom can be empty, default to time.Now()

Issue #3, opened 2017-02-10
Create Istio cert and key for Kubernetes' service account


lookuptable:
Per the design of Istio auth, we should monitor the creation of Kubernetes service account and automatically create a key and certificate pair by calling into the certificate issuance API ( #1 ). The key and certificate are used to establish secure connection/channel between two Istio proxies.

lookuptable:
#38 closes this issue.

PR #27, opened 2017-02-28
Add SecretController that watches service accounts to create key & cert

lookuptable:
This is the first commit towards #3 .

### DIFF
--- a/controller/secret.go
+++ b/controller/secret.go
@@ -1,3 +1,17 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


controller/secret.go
```diff
@@ -0,0 +1,149 @@
...
+	acct := obj.(*v1.ServiceAccount)
+	sc.deleteSecret(acct.GetName(), acct.GetNamespace())
+}
+
+func (sc *SecretController) updateFunc(oldObj, curObj interface{}) {
+	if reflect.DeepEqual(oldObj, curObj) {
+		// Nothing is changed. The method is invoked by periodical re-sync with the apiserver.
+		return
+	}
+	oldSa := oldObj.(*v1.ServiceAccount)
```
wattli: Should we make sure oldObj is not null?

lookuptable: I don't think it's necessary. The `AddFunc` method should be called when oldObj is nil

PR #36, opened 2017-03-07
Use URI type for Isito identities

lookuptable:


certmanager/generate_cert.go
```diff
@@ -174,18 +189,40 @@ func genCertTemplate(options CertOptions) x509.Certificate {
 		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
 		ExtKeyUsage:           []x509.ExtKeyUsage{extKeyUsage},
 		BasicConstraintsValid: true,
-	}
-	hosts := strings.Split(options.Host, ",")
```
wlu2016: shall we keep these options (DNS and IPaddress)? we might need that for testing when we work on enduser to TLS-secured server

certmanager/generate_cert.go
```diff
@@ -174,18 +189,40 @@ func genCertTemplate(options CertOptions) x509.Certificate {
...
+		ExtraExtensions:       []pkix.Extension{sanExt},
 	}
 	if options.IsCA {
 		template.IsCA = true
 		template.KeyUsage |= x509.KeyUsageCertSign
 	}
 	return template
 }
+
+func buildSubjectAltNameExtension(host string) pkix.Extension {
```
wlu2016: also update https://github.com/istio/auth/tree/master/cmd/generate_cert to use URI?

certmanager/generate_cert.go
```diff
@@ -66,8 +67,21 @@ type CertOptions struct {
...
+	// Size of RSA key to generate.
+	rsaBits = 2048
+
+	// OID tag values for X.509 SAN field (see https://tools.ietf.org/html/rfc5280#appendix-A.2)
+	tagDNSName = 2
+	tagURI     = 6
+	tarIP      = 7
+
+	// Prefix for Istio identities.
+	prefixIstioID = "istio:"
```
wlu2016: prefixIstioID -> uriScheme

certmanager/generate_cert.go
lookuptable: Those options are still supported. The logic in Line 205-218 is:
```
if isIp():
  Use IP address
elsif host begins with "istio:"
  Use URI
else
  Use DNS
```

certmanager/generate_cert.go
lookuptable: We don't need to update there. It will automatically choose URI or not depending on the host string prefix.

certmanager/generate_cert.go
lookuptable: Done

certmanager/generate_cert.go
```diff
@@ -66,8 +67,21 @@ type CertOptions struct {
...
+	tagDNSName = 2
+	tagURI     = 6
+	tarIP      = 7
+
+	// The URI scheme for Istio identities.
+	uriScheme = "istio:"
+)
+
+// See http://www.alvestrand.no/objectid/2.5.29.17.html.
+var oidSubjectAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
```
lizan: const?

lookuptable: Golang does not allow this to be defined as a const var since the object is not immutable

PR #38, opened 2017-03-09
Mint Istio identities

lookuptable:


certmanager/ca.go
```diff
@@ -14,7 +14,71 @@
...
+	"crypto/x509"
+	"fmt"
+	"time"
+)
+
+const (
+	// The time to live for issued certificate.
+	certTTL = time.Hour
+	// The time to live for self-signed root CA certificate.
+	rootCertTTL = 24 * time.Hour
```
lizan: where does these numbers coming from?

certmanager/ca.go
```diff
@@ -14,7 +14,71 @@
...
 // CertificateAuthority contains methods to be supported by a CA.
 type CertificateAuthority interface {
-	Generate(name string) (key, cert []byte)
+	Generate(name, namespace string) (cert, key []byte)
+}
+
+// IstioCA generates keys and certificates for Istio identities.
+type IstioCA struct {
+	signerCert *x509.Certificate
+	signerKey  *rsa.PrivateKey
```
lizan: use `crypto.PrivateKey` so the key doesn't have to be RSA (it could be ECDSA)? or we can just mark it for now we only support RSA keys.

certmanager/ca.go
lookuptable: Just some numbers that sound reasonable to me (unless you disagree). Later these should be configurable with these as defaults.

certmanager/ca.go
lookuptable: Done!

certmanager/util.go
```diff
@@ -0,0 +1,47 @@
...
+		glog.Fatalf("Invalid PEM encoding: %s", pb)
+	}
+	cert, err := x509.ParseCertificate(b.Bytes)
+	if err != nil {
+		glog.Fatalf("Failed to parse X.509 certificate (error: %s)", err)
+	}
+	return cert
+}
+
+func parsePemEncodedPrivateKey(pb []byte) *rsa.PrivateKey {
```
lizan: support other private key algorithms here too?

lookuptable: Any suggestion on how to do that? Golang x509 do not have a method to parse any key so we need to trial-and-error, which doesn't look very clean to me.

lizan: https://golang.org/pkg/crypto/x509/#PublicKeyAlgorithm
Certificate has PublicKeyAlogrithm, so it should match privatekey algorithm.

lookuptable: Done. PTAL @lizan

PR #39, opened 2017-03-09
Add test descriptions for securenaming

lookuptable:


controller/securenaming_test.go
```diff
@@ -94,7 +90,7 @@ func TestProcessNextServices(t *testing.T) {
 		},
 	}
 
-	for i, c := range cases {
+	for d, c := range cases {
```
wlu2016: why changing i to d? is this a new convention? 'i' seems more popular to me

lookuptable: Because it's no longer an integer index. It's now a string description for the test case.

wlu2016: lgtm

PR #40, opened 2017-03-09
Add a script to detect whether go files contain copyright headers

### LINKED ISSUE #33: Check copyright header in Travis
We should check the copyright header in Travis CI to avoid accidentally committing new files without the headers.

lookuptable:
This closes #33 .

### DIFF
--- a/bin/headers.sh
+++ b/bin/headers.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+failure=0
+for file in $(find . -name "*.go"); do
+  head -n 1 $file | grep '^// Copyright 2017 Istio Authors' > /dev/null
+  if [[ $? -ne 0 ]]; then
+    echo $file does not have a copyright license header
+    failure=1
+  fi
+done
+
+exit $failure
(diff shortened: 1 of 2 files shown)


bin/headers.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+failure=0
```
wlu2016: failure -> failures

bin/headers.sh
```diff
@@ -0,0 +1,12 @@
...
+for file in $(find . -name "*.go"); do
+  head -n 1 $file | grep '^// Copyright 2017 Istio Authors' > /dev/null
```
wlu2016: remove "Copyright 2017"? or use a regex to match "Copyright any_year Istio Authors"

bin/headers.sh
lookuptable: This is a flag indicating whether there's any error. I'll change it to something like `has_failure` to make it more meaningful.

bin/headers.sh
lookuptable: Done

PR #44, opened 2017-03-14
Add cert sign key usage when the certificate is self-signed

lookuptable:


certmanager/generate_cert.go
```diff
@@ -166,19 +166,27 @@ func genSerialNum() *big.Int {
 // genCertTemplate generates a certificate template with the given options.
 func genCertTemplate(options CertOptions) x509.Certificate {
 	notBefore, notAfter := toFromDates(options.ValidFrom, options.ValidFor)
+
+	keyUsage := x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment
```
wlu2016: add a comment about these two KeyUsage for?

certmanager/generate_cert.go
```diff
@@ -166,19 +166,27 @@ func genSerialNum() *big.Int {
...
+	if options.IsSelfSigned {
+		keyUsage |= x509.KeyUsageCertSign
```
wlu2016: add comment why self signed key needs KeyUsageCertSign?

PR #53, opened 2017-03-16
Fix flakey tests in securenaming

### LINKED ISSUE #28: Investigate test flakeness
Currently `TestGetPodServices` fails occasionally. We should investigate and fix this flakiness.

lookuptable:
This closes #28 .

### DIFF
--- a/controller/securenaming_test.go
+++ b/controller/securenaming_test.go
@@ -199,10 +199,7 @@ func TestGetPodServices(t *testing.T) {
 		actualServices := snc.getPodServices(testCase.pod)
 		sort.Sort(sortableServices(actualServices))
 
-		expectedServices := testCase.expectedServices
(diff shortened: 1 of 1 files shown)


controller/securenaming_test.go
```diff
@@ -195,8 +197,12 @@ func TestGetPodServices(t *testing.T) {
 		}
 
 		actualServices := snc.getPodServices(testCase.pod)
+		sort.Sort(sortableServices(actualServices))
 
-		if !reflect.DeepEqual(actualServices, testCase.expectedServices) {
+		expectedServices := testCase.expectedServices
+		sort.Sort(sortableServices(expectedServices))
```
lizan: I believe expectedServices is already sorted, so no need Sort again?

lookuptable: Good point. Done.

PR #55, opened 2017-03-17
Make host optional

### LINKED ISSUE #46: host in CertOptions should be optional
We should make `CertOptions.Host` optional to avoid setting SAN field when generating CA certificates.

lookuptable:
This allows us to generate root/intermediate CA certicates.

Fix #46


wattli:
Are you guys crazy :)

PR #58, opened 2017-03-17
[WIP] Istio CA should return a cert chain instead of a single cert

lookuptable:


certmanager/ca.go
```diff
@@ -50,22 +63,36 @@ func NewSelfSignedIstioCA() *IstioCA {
...
+	ca.signingKey = parsePemEncodedKey(ca.signingCert.PublicKeyAlgorithm, opts.SigningKeyBytes)
+
+	if err := ca.verify(); err != nil {
+		return nil, err
 	}
+
+	return ca, nil
 }
 
 // Generate returns a key and a certificate of an Istio identity defined by
```
lookuptable: `returns a certificate chain and a key`

PR #59, opened 2017-03-17
Istio CA should return a cert chain instead of a single cert

### LINKED ISSUE #56: Handle key/certificate rotations

lookuptable:
Fix #56


certmanager/ca.go
```diff
@@ -50,22 +63,36 @@ func NewSelfSignedIstioCA() *IstioCA {
 		IsSelfSigned: true,
 	}
 	pemCert, pemKey := GenCert(options)
-	cert, key := parsePemEncodedCertificateAndKey(pemCert, pemKey)
 
-	return NewIstioCA(cert, key)
+	opts := &IstioCAOptions{
+		CertChainBytes:   pemCert,
```
lizan: In self-signed root certificate case, you don't need to include the root cert here.

Basically the only certificates needed to be included are the one before root certificate, since the root certificate is already distributed to the other end.

https://tools.ietf.org/html/rfc5246#section-7.4.2

It is just an optimization, it is ok to include, but not necessary.

certmanager/ca_test.go
```diff
@@ -49,3 +67,96 @@ func TestSelfSignedIstioCA(t *testing.T) {
 		t.Errorf("Generated certificate does not contain a SAN field")
 	}
 }
+
+// Pass in unmatched chain and cert to make sure the `verify` method yeilds an error.
+func TestInvalidIstioCAOptions(t *testing.T) {
+	rootCert := `
+-----BEGIN CERTIFICATE-----
```
wattli: Is there any timestamp in the cert? If so, would it still work when the cert expires?

certmanager/ca.go
lookuptable: Done. We are on longer sending root cert in the cert chain in self-signed case. 

In non-self-signed case we have no control over what's in the chain file. So we'll blindly send over what's in the file.

certmanager/ca_test.go
lookuptable: Updated the timestamps to make the certs long-lived so expiration shouldn't be an issue.

wattli:
lgtm

PR #60, opened 2017-03-18
Use glog instead of log to stay consistent across repo

lookuptable:


wattli:
please fix failed test.

PR #62, opened 2017-03-21
Allow certificate to be used as both client-side and server-side cert

lookuptable:


certmanager/ca.go
```diff
@@ -103,7 +103,9 @@ func (ca IstioCA) Generate(name, namepsace string) (chain, key []byte) {
 		SignerCert:   ca.signingCert,
 		SignerPriv:   ca.signingKey,
 		IsCA:         false,
+		IsClient:     true,
```
wattli: Do we need these options (IsClient and IsServer) any more? Since CA can simply generate a generic server used by server and client.

lookuptable: I kind like keeping them - the CA cert doesn't need to set IsClient and IsServer. I may clean more things up in a followup PR.

PR #64, opened 2017-03-22
Better handling existing secret

### LINKED ISSUE #48: Better handling for existing secrets
Currently we use `Update` instead of `Create` to avoid dealing with existing secrets. However we need to handle this more properly.

Relevant code is here: https://github.com/istio/auth/blob/master/controller/secret.go#L121

lookuptable:
Fix #48

### DIFF
--- a/controller/secret.go
+++ b/controller/secret.go
@@ -74,7 +74,8 @@ func NewSecretController(ca certmanager.CertificateAuthority, core corev1.CoreV1
 			return core.Secrets(metav1.NamespaceAll).Watch(options)
 		},
 	}
-	c.scrtStore, c.scrtController = cache.NewInformer(secretlw, &v1.Secret{}, time.Minute, cache.ResourceEventHandlerFuncs{})
+	c.scrtStore, c.scrtController =
+		cache.NewInformer(secretlw, &v1.Secret{}, time.Minute, cache.ResourceEventHandlerFuncs{})
 
 	return c
 }
--- a/controller/secret.go
+++ b/controller/secret.go
@@ -36,10 +36,12 @@ const secretNamePrefix = "istio."
 
 // SecretController manages the service accounts' secrets that contains Istio keys and certificates.
 type SecretController struct {
-	ca           certmanager.CertificateAuthority
-	core         corev1.CoreV1Interface
-	saController cache.Controller
-	saStore      cache.Store
+	ca             certmanager.CertificateAuthority
+	core           corev1.CoreV1Interface
+	saController   cache.Controller
+	saStore        cache.Store
+	scrtController cache.Controller
+	scrtStore      cache.Store
 }
 
 // NewSecretController returns a pointer to a newly constructed SecretController instance.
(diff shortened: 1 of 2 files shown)


controller/secret.go
```diff
@@ -36,10 +36,12 @@ const secretNamePrefix = "istio."
...
 type SecretController struct {
-	ca           certmanager.CertificateAuthority
-	core         corev1.CoreV1Interface
-	saController cache.Controller
-	saStore      cache.Store
+	ca             certmanager.CertificateAuthority
+	core           corev1.CoreV1Interface
+	saController   cache.Controller
+	saStore        cache.Store
+	scrtController cache.Controller
```
wattli: Add some comments about this variable and the one below?

controller/secret.go
```diff
@@ -62,20 +64,31 @@ func NewSecretController(ca certmanager.CertificateAuthority, core corev1.CoreV1
...
 func (sc *SecretController) Run(stopCh chan struct{}) {
+	go sc.scrtController.Run(stopCh)
 	go sc.saController.Run(stopCh)
 	<-stopCh
 }
 
 func (sc *SecretController) addFunc(obj interface{}) {
 	acct := obj.(*v1.ServiceAccount)
-	sc.createSecret(acct.GetName(), acct.GetNamespace())
+	sc.createOrUpdateSecret(acct.GetName(), acct.GetNamespace())
```
wattli: Nit: maybe call it upsertSecret?

controller/secret.go
wattli: Or just call it 'secretController' to be consistent with secretlw below.

controller/secret.go
```diff
@@ -49,7 +51,7 @@ func NewSecretController(ca certmanager.CertificateAuthority, core corev1.CoreV1
 		core: core,
 	}
 
-	lw := &cache.ListWatch{
+	salw := &cache.ListWatch{
```
wattli: saLw? To make it more readable. Same for secretLw.

controller/secret.go
```diff
@@ -62,20 +64,31 @@ func NewSecretController(ca certmanager.CertificateAuthority, core corev1.CoreV1
...
+
+	secretlw := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			return core.Secrets(metav1.NamespaceAll).List(options)
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return core.Secrets(metav1.NamespaceAll).Watch(options)
+		},
+	}
+	c.scrtStore, c.scrtController = cache.NewInformer(secretlw, &v1.Secret{}, time.Minute, cache.ResourceEventHandlerFuncs{})
```
wattli: So you will have a followup PR for this handlerFuncs? If so, please leave a TODO here.

controller/secret.go
lookuptable: Done

controller/secret.go
lookuptable: Done

controller/secret.go
lookuptable: Done

PR #75, opened 2017-03-27
Auto refreshing secrets before the contained cert expires

### LINKED ISSUE #56: Handle key/certificate rotations

lookuptable:
Fix #56

### DIFF
--- a/controller/secret.go
+++ b/controller/secret.go
@@ -34,8 +34,15 @@ import (
 
 /* #nosec: disable gas linter */
 const (
-	secretNamePrefix = "istio."
-	istioSecretType  = "istio.io/key-and-cert"
+	secretNamePrefix   = "istio."
+	istioSecretType    = "istio.io/key-and-cert"
+	secretResyncPeriod = time.Minute
+
+	serviceAccountNameAnnotationKey = "istio.io/service-account.name"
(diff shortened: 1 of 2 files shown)


controller/secret.go
```diff
@@ -149,18 +159,17 @@ func (sc *SecretController) upsertSecret(saName, saNamespace string) {
...
 	}
 
 	// Now we know the secret does not exist yet. So we create a new one.
 	chain, key := sc.ca.Generate(saName, saNamespace)
 	rootCert := sc.ca.GetRootCertificate()
 	secret.Data = map[string][]byte{
-		"cert-chain.pem": chain,
-		"key.pem":        key,
-		"root-cert.pem":  rootCert,
+		certChainSecretKey:  chain,
```
wattli: The trailing 'Key' makes the variable misleading (I thought it's a key instead of a cert). What about call it certChainSecret or something?

wattli: Same for rootCertSecretKey.

lookuptable: I was following the convention used in kubernetes code base but I can see the confusion. I guess we can rename it to `rootCertSecreteMapKey` to make it clear that it's a key in a map.

Issue #76, opened 2017-04-03
Move presubmit to Jenkins


yutongz:
Manager, proxy and other repos even the postsubmit of auth use Jenkins, it would be nice if we can move presubmit of auth into Jenkins. It makes sharing scripts and tools more easily.

yutongz:
I am working on this issue

lookuptable:
Thanks for volunteering!

yutongz:
Fixed with https://github.com/istio/auth/pull/77

PR #77, opened 2017-04-04
Add presubmit step on Jenkins

yutongz:
https://github.com/istio/auth/issues/76


Jenkinsfile
```diff
@@ -32,3 +36,29 @@ def postsubmit(gitUtils, bazel, utils) {
     }
   }
 }
+
+def presubmit(gitUtils, bazel, utils) {
+  goBuildNode(gitUtils, 'istio.io/mixer') {
```
lookuptable: istio.io/auth

bin/install-prereqs.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Install linters
+go get -u github.com/alecthomas/gometalinter
+gometalinter --install --vendored-linters
+
+# Install buildifier BUILD file validator
+go get -u github.com/bazelbuild/buildifier/buildifier
```
lookuptable: is buildifier used any where?

Jenkinsfile
```diff
@@ -32,3 +36,29 @@ def postsubmit(gitUtils, bazel, utils) {
...
+    stage('Bazel Tests') {
+      bazel.test('//...')
+    }
+    stage('Code Check') {
+      sh('bin/linters.sh')
+      sh('bin/headers.sh')
+    }
+    stage('Code Coverage') {
+      sh('bin/codecov.sh')
+      utils.publishCodeCoverage('AUTH_CODECOV_TOKEN')
```
lookuptable: Is this token defined somewhere?

yutongz: Would you like to go and check if you can add this token? 
https://docs.codecov.io/v4.3.0/reference#usage

Jenkinsfile
yutongz: Oh!

bin/install-prereqs.sh
yutongz: It will be used in another check file. We want to add that check for auth later. Well, I can remove it since it's not used right now, I guess.

lookuptable: yeah let's add it when it actually gets used to keep each commit self-contained.

sebastienvas:
jenkins please rebuild auth/presubmit

PR #84, opened 2017-04-07
Add package code coverage check

yutongz:
https://github.com/istio/istio-testing/issues/135
Add a step to check if the code coverage meets requirement in every specific package.

### DIFF
--- a/bin/toolbox/presubmit/package_coverage_check.py
+++ b/bin/toolbox/presubmit/package_coverage_check.py
@@ -1,0 +1,15 @@
+from istio_toolbox.pkgCheck import pkg_cc_check
+import argparse
+import sys
+
+def packageCodeCoverageCheck(report, requirement):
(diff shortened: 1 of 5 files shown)


bin/toolbox/presubmit/package_coverage_check.py
```diff
@@ -0,0 +1,15 @@
+from istio_toolbox.pkgCheck import pkg_cc_check
```
lookuptable: Just out of curiosity: why not put this file in the toolbox repo? It seems to be repo-independent?

yutongz: Yes, you are right, it's definitely repo-independent. However, I just can't find a direct way to call a program from another repo. That's why I put this small part of code as a entry such that I can call the really function in toolbox repo.

PR #90, opened 2017-04-13
Allow specifying cert TTL via command line flags

lookuptable:


cmd/istio_ca/main.go
```diff
@@ -46,6 +47,11 @@ var (
 	selfSignedCA = flag.Bool("self-signed-ca", false,
 		"Indicates whether to use auto-generated self-signed CA certificate. "+
 			"When set to true, the '-ca-cert' and '-ca-key' options are ignored.")
+
+	caCertTTL = flag.Duration("ca-cert-ttl", 240*time.Hour,
```
wattli: Just curious, what will happen when caCert expires while proxy cert is still valid? Would mTLS break?

lookuptable: mTLS will break since the whole cert chain needs to be valid.

wattli: As discussed offline, this does not seem to be an issue for alpha. We probably won't have time to support cert rotation considering the code complete date is tomorrow.

PR #93, opened 2017-04-26
Update Istio secret when root cert is changed

lookuptable:


controller/secret.go
```diff
@@ -204,19 +205,29 @@ func (sc *SecretController) scrtUpdated(oldObj, newObj interface{}) {
...
-		glog.Infof("Certificate %s is about to expire, refreshing it", scrt.GetName())
+	rootCertificate := sc.ca.GetRootCertificate()
+
+	// Refresh the secret if 1) the certificate contained in the secret is about
+	// to expire, or 2) the root certificate in the secret is different than the
+	// one held by the certmanager (this may happen when the CA is restarted and
+	// a new self-signed CA cert is generated).
+	if ttl.Seconds() < secretResyncPeriod.Seconds() || !bytes.Equal(rootCertificate, scrt.Data[rootCertID]) {
+		namespace := scrt.GetNamespace()
+		name := scrt.GetName()
```
lookuptable: The name gets logged is the secret name.

PR #101, opened 2017-05-11
Update auth for qualification to follow same pattern as mixer and manager

sebastienvas:
This will ease out release.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library('testutils@stable-cd138c4')
+@Library('testutils@stable-41b0bf6')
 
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -60,3 +58,45 @@ def presubmit(gitUtils, bazel, utils) {
     }
   }
 }
+
+def postsubmit(gitUtils, bazel, utils) {
+  goBuildNode(gitUtils, 'istio.io/mixer') {
```
lookuptable: istio.io/auth

sebastienvas: thanks. Fixed

PR #136, opened 2017-05-24
Re-format README.md

lookuptable:
Mostly about wrapping up long lines

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,17 +5,29 @@
 
 ## Overview
 
-Istio Auth aims at enhancing the security of microservices and their communication without requiring service code changes. It is responsible for:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,74 +5,152 @@
...
+photo-frontend cannot access datastore.
+
+In this setup, Istio CA is able to provide key/cert management for all
+namespaces. And we successfully prevent the team from messing up services
+running by other teams.
 
 ## Future Work
+
 - Fine-grained authorization and auditing
 - Secure Istio components (mixer, discovery service, etc.)
```
wattli: Should we call this Istio Mixer and Istio Manager, to be consistent with the ones above?

lookuptable: Yes, we should. Another PR is coming up!

PR #137, opened 2017-05-24
Update the names of Istio components

lookuptable:


README.md
```diff
@@ -152,7 +152,7 @@ running by other teams.
 ## Future Work
 
 - Fine-grained authorization and auditing
-- Secure Istio components (mixer, discovery service, etc.)
+- Secure Istio components (Mixer, Istio-Manager, etc.)
```
wattli: How about 

Istio Mixer, Istio Manager, etc.

to be consistent with the terms used previously in this doc.

lookuptable: I'm following the conventions here: https://istio.io/docs/concepts/what-is-istio/overview.html. There it has Mixer, Istio-Manager

README.md
```diff
@@ -92,7 +92,7 @@ cover each phase in this section and a more detailed version can be found
 2. When a pod is created, API Server mounts the cert/key according to the
    service account using [Kubernetes
    secrets](https://kubernetes.io/docs/concepts/configuration/secret/).
-3. [Istio Manager](https://github.com/istio/manager/blob/master/doc/design.md)
+3. [Istio-Manager](https://github.com/istio/manager/blob/master/doc/design.md)
```
wattli: This is an interesting naming convention. https://screenshot.googleplex.com/YF1B3pPoq5T, not sure if we should change it for Auth as well. I will let you decide.

PR #162, opened 2017-06-27
update test-infra version to upload coverage to GCS

yutongz:
1. Update test-infra version, for uploading coverage report to GCS  https://github.com/istio/test-infra/issues/221
2. Rename coverage.sh to codecov.sh for consistence


yutongz:
@wattli

PR #164, opened 2017-06-29
Istio CA Service definition

sarvaniv:
First draft of Istio CA service definition.


sarvaniv:
@wattli @myidpt @lookuptable Please take a look and comment :-)

proto/ca_service.pb.go
```diff
@@ -0,0 +1,316 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// source: ca_service.proto
```
wattli: Why do we need this generated file?

proto/ca_service.proto
```diff
@@ -0,0 +1,63 @@
...
+    OnPremNodeCredential on_prem_node_credential = 2;
+  }
+}
+
+message GcpInstanceCredential {
+  bytes signed_instance_metadata = 1;
+}
+
+message OnPremNodeCredential {
+  bytes node_identity_cert = 1;
```
myidpt: I think you can just have node_identity_cert_chain which includes the entire cert chain. This is how Envoy is configured and I suppose it's also how OpenSSL takes the certs.

proto/ca_service.proto
```diff
@@ -0,0 +1,63 @@
...
+    GcpInstanceCredential gcp_instance_credential = 1;
+    OnPremNodeCredential on_prem_node_credential = 2;
+  }
+}
+
+message GcpInstanceCredential {
+  bytes signed_instance_metadata = 1;
+}
+
+message OnPremNodeCredential {
```
myidpt: Actually I'm not sure you should have this. The mTLS should be established (and the node agent cert is passed over) before the real data transmission.

wattli: +1, client cert should be passed in the channel, not the request.

wattli: http://krishicks.com/post/grpc-mutual-tls-golang/#server-configuration This is a good doc, FYI.

proto/ca_service.proto
```diff
@@ -0,0 +1,63 @@
...
+syntax = "proto3";
+
+package proto;
+
+// Service definition of IstioCAService that can dynamically verify the CSR for
+// service identities for Istio services running on various platforms and
+// approve or disapprove the request after verifying the Node Agent credentials
+// provided as a part of the Certificte Signing Request.
+service IstioClusterCAService {
+
```
lookuptable: Should this be IstioCAService'

proto/ca_service.proto
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+message GcpInstanceCredential {
+  bytes signed_instance_metadata = 1;
+}
+
+message OnPremNodeCredential {
+  bytes node_identity_cert = 1;
+  bytes cert_chain = 2;
```
lookuptable: Why not put node_identity_cert into the chain?

proto/ca_service.proto
sarvaniv: The reason I added this is because I need to get the node identity from the cert. While we need to provide the identity on the channel, I will also need the identity on the cert to get the Node Identity so I can get the service identity for the node. I provided both the individual as well as cert chain since I was not fully sure if it is possible to provide the certs in both ways - i.e., individual cert and cert chain. Another option is for the Node Agent to provide the identity but I am not too sure if this is something the Node Agent should do. It just provides the cert / cert-chain to the CA and the CA will sign the CSR with the service identity.

proto/ca_service.proto
sarvaniv: I am calling out the "Cluster" for a reason. An Istio mesh can be made up of multiple clusters but each cluster has a CA. We don't have a single CA for the entire mesh. Does that answer your question?

proto/ca_service.proto
sarvaniv: Sure. That can be done. I was not sure how the certs are being provided right now. May be we can get away with just the cert chain.

proto/ca_service.proto
myidpt: I don't think cluster is the right term here. If you have a single CA for multiple K8s clusters, it is strange to call that CA a cluster CA. In one K8s cluster we may also have multiple namespace-scoped CAs. Cluster is a over-used term. If you don't clearly define what cluster is, people have many different ways to interpret it.

proto/ca_service.proto
myidpt: Yes. That's the way Envoy is configured.

proto/ca_service.proto
sarvaniv: When we are talking about Istio on hybrid clusters what is the composition of each cluster in the hybrid cluster? I don't think cluster is over-used. I think it is used more to describe a set of nodes that can work together in isolation. And it is used so much because it is better understood in the community.  Note that the service definition is only for the non-k8s cluster cases. In this case, we have an Istio cluster that will have its own CA, its own Pilot an ingress and egress proxy.

proto/ca_service.proto
sarvaniv: Changed to just have cert chain.

proto/ca_service.proto
wattli: You can get node identity from client cert in the mTLS connection. It's in the example I sent to you last night. You can't rely on cert in request.

sarvaniv: I see. Removed.

sarvaniv: I thought I would verify the cert once again on CA or just verify it on CA. But if we get get the identity, then that is good.

proto/ca_service.proto
```diff
@@ -0,0 +1,57 @@
...
+  rpc Sign(CertificateSignRequest) returns (CertificateSignResponse);
+}
+
+message CertificateSignRequest {
+  bytes csr = 1;
+  NodeAgentCredentials nodeAgentCredentials = 2;
+}
+
+message NodeAgentCredentials {
+  oneof node_credential {
```
wattli: This does not make sense. Remove?

proto/ca_service.proto
```diff
@@ -0,0 +1,57 @@
...
+  bytes signed_instance_metadata = 1;
+}
+
+message CertificateSignResponse {
+  bool is_approved = 1;
+  Error error = 2;
+  bytes signed_cert_chain = 3;
+}
+
+message Error {
```
wattli: There is a standard: https://godoc.org/google.golang.org/grpc/codes

proto/ca_service.proto
sarvaniv: Why? For now, we are only handling GCP. But later we could handle another cloud provider like AWS. In that case, it will have another entry.

proto/ca_service.proto
lookuptable: @sarvaniv what do you mean by saying 'the service definition is only for the non-k8s cluster cases'. I'd prefer the CA works for both VM and K8S eventually.

Also in the current definition, we are not enforcing the 'cluster' constraint anywhere. Thus having cluster in the name is more confusing than explanatory.

proto/ca_service.proto
sarvaniv: This code is more for the application and not at grpc level. I will look into whether this can be directly used but the code maps to a specific grpc error which is not necessarily the same at the CA application generated error.

proto/ca_service.proto
sarvaniv: Cluster is more a logical concept than a physical concept. The naming is to provide the intended behavior of the module rather than enforce constraints. You can look at https://docs.google.com/document/d/1ab_f7OnsuQ3Q8C-Al6yHr1pFNv_Z8BtfovzYEU5dII4/edit to see that a cluster is being using to define an Istio mesh on a Hybrid cloud.

proto/ca_service.proto
sarvaniv: Also, this is in golang and not in protobuf. I don't think I can use it here in the proto file.

lizan: Use google/rpc/status.proto https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto

This is consistent across all istio apis.

proto/ca_service.proto
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+message GcpInstanceCredential {
+  bytes signed_instance_metadata = 1;
+}
+
+message CertificateSignResponse {
+  bool is_approved = 1;
+  Error error = 2;
```
lizan: You don't need this, just use gRPC error codes.

proto/ca_service.proto
```diff
@@ -0,0 +1,57 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package proto;
```
lizan: package istio.v1.auth

wattli:
@geeknoid , this protobuf is used by istio node agent calling istio CA (they are both istio auth components). Should we put this in istio/api or this repo?

geeknoid:
​If the node agent and Istio CA are in different repos, then put the API in
istio/api. If they're both in the same repo, then we consider the interface
to be private and so can go in the repo itself.​

PR #165, opened 2017-06-30
Add generate csr function

wattli:


cmd/generate_csr/main.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017 Istio Authors
```
lookuptable: Just curious: who is going to call this executable?

certmanager/generate_cert.go
```diff
@@ -142,6 +168,22 @@ func genSerialNum() *big.Int {
 	return serialNum
 }
 
+// genCertTemplate generates a certificateRequest template with the given options.
```
lookuptable: s/genCertTemplate/genCsrTemplate

certmanager/generate_cert.go
```diff
@@ -83,6 +83,32 @@ const (
 // See http://www.alvestrand.no/objectid/2.5.29.17.html.
 var oidSubjectAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
 
+// GenCSR generates a X.509 certificate sign request with the given options.
```
lookuptable: This method returns both a CSR and the private key. Can you update the comment?

I notice we have a similar issue on line 112. I'll fix that later

certmanager/generate_cert.go
```diff
@@ -83,6 +83,32 @@ const (
...
+func GenCsr(options CertOptions) ([]byte, []byte) {
```
lookuptable: s/GenCsr/GenCSR

certmanager/generate_cert.go
```diff
@@ -83,6 +83,32 @@ const (
...
+	// Generates a CSR
+	// The public key will be bound to the certficate generated below. The
+	// private key will be used to sign this certificate in the self-signed
+	// case, otherwise the certificate is signed by the signer private key
+	// as specified in the CertOptions.
```
lookuptable: Some of the comments are not appropriate here - this method should only generate a CSR and not worry about who is going to sign it.

certmanager/generate_cert.go
```diff
@@ -83,6 +83,32 @@ const (
...
+		glog.Fatalf("RSA key generation failed with error %s.", err)
+	}
+	template := genCsrTemplate(options)
+	csrBytes, err := x509.CreateCertificateRequest(rand.Reader, &template, crypto.PrivateKey(priv))
+	if err != nil {
+		glog.Fatalf("Could not create certificate (err = %s).", err)
+	}
+
+	// Returns the certificate that carries the RSA public key as well as
+	// the corresponding private key.
```
sarvaniv: The comment says the certificate carries RSA public key as well as private key. This does not sound right to me. The CSR should contain a digital signature signed with the private key. Could you clarify?

cmd/generate_csr/main.go
wattli: The node agent that lives in node, see https://docs.google.com/document/d/1RZnBfj8I5xs8Yi_BPEKBRp0K3UnIJYTDg_31rfTt4j8/edit#heading=h.p0zc37bjhmkg

certmanager/generate_cert.go
wattli: Both are updated since i am already here.

certmanager/generate_cert.go
wattli: Updated.

certmanager/generate_cert.go
wattli: Done

certmanager/generate_cert.go
sarvaniv: nice :-)

certmanager/generate_cert.go
wattli: Reworded. It actually means returns a CSR and a private key

certmanager/generate_cert.go
sarvaniv: As discussed, lets not replicate code. Some refactoring will help keep common code in a function.

wattli: done

certmanager/generate_cert_test.go
```diff
@@ -36,6 +36,36 @@ type VerifyFields struct {
 
 var now = time.Now().Round(time.Second).UTC()
 
+func TestGenCsr(t *testing.T) {
```
sarvaniv: Do we need tests for negative case? For example, invalid csr options?

PR #168, opened 2017-07-07
Add util function to get identity token from VM

wattli:


certmanager/util.go
```diff
@@ -18,10 +18,40 @@ import (
...
 )
 
+const (
+	// The GCE metadata service URI to get identity token.
+	identityUri = "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity"
+)
+
+// Get the GCE VM identity jwt token from its metadata server.
+// Note: this function only works in a GCE VM environment.
+func GetIdentityToken(aud string) ([]byte, error) {
```
sarvaniv: Could you instead have an interface as follows:

type TokenFetcher interface {
       TokenFetch(...)
}

type GcpTokenFetcher struct {
         const identityUri = "http://...."
}

func TokenFetch() ([]byte, error) {
}

This will make it very easy to incorporate a new identity to be implemented and tested.

lizan: what is expected audience format? Need a comment or a link.

certmanager/util.go
```diff
@@ -18,10 +18,40 @@ import (
...
+	req, err := http.NewRequest("GET", identityUri+aud, nil)
```
lizan: identityUri + "?audience=" + aud?

certmanager/util.go
sarvaniv: I don't think the code is testable in the current form. Can we mock the request and response so we can test the code?

lookuptable: I don't feel `util.go` in `certmanager` package is the right place for this method. It will be confusing when reading `certmanager.GetIdentityToken()` as this method only returns the default ID token in GCP environment.

Maybe we can structure the code in a folder like `pkg/providers/gcp` and call the method as `gcp.GetDefaultIDToken()`. This allows us to have a similar support for non-GCP environment like aws

certmanager/util.go
```diff
@@ -18,10 +18,40 @@ import (
...
+	if err != nil {
+		glog.Errorf("Fail to construct the http request: %s", err)
+		return nil, err
+	}
+	req.Header.Set("Metadata-Flavor", "Google")
```
lookuptable: Can we utilize anything from https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/compute/metadata/metadata.go to avoid doing these low-level stuff?

certmanager/util.go
```diff
@@ -18,10 +18,40 @@ import (
 	"crypto"
 	"crypto/x509"
 	"encoding/pem"
+	"net/http"
 
 	"github.com/golang/glog"
 )
 
+const (
+	// The GCE metadata service URI to get identity token.
```
lookuptable: get ID token for the default service account?

certmanager/util.go
sarvaniv: While I think this is a really good idea, I don't think the code is returning a signed metadata but I may be wrong. My understanding is that the signed metadata is GA only last week or so. Double check before you use it.

certmanager/util.go
wattli: more comments are added as well as the original link.

certmanager/util.go
wattli: good point, i put it at the end of identityUri instead.

certmanager/util.go
wattli: Make sense, done.

wattli: This func is fairly simple, only calls two util function and do basic err check, so it's not trivial to do the unit test. Can we skip this one since I tested it manually in a GCE.

certmanager/util.go
wattli: Right, the url is fixed even for non-default service account.

lookuptable: That's not entirely true, no?

To get token for SA 'foo', the URL needs to be service-accounts/foo/identity

certmanager/util.go
wattli: It actually works, thanks Yang.

certmanager/util.go
```diff
@@ -18,10 +18,43 @@ import (
...
+}
+
+type GcpTokenFetcher struct {
+	// aud is the unique URI agreed upon by both the instance and the system verifying the instance's identity.
+	// For more info: https://cloud.google.com/compute/docs/instances/verifying-instance-identity
+	aud            string
+	serviceAccount string
+}
+
+func (fetcher GcpTokenFetcher) setAudience(audience string) {
```
sarvaniv: Nit: Make this a pointer. (fetcher *GcpTokenFetcher)

certmanager/util.go
```diff
@@ -18,10 +18,43 @@ import (
 	"crypto"
 	"crypto/x509"
 	"encoding/pem"
+	"net/http"
 
+	// Install this package by $ go get -u cloud.google.com/go/...
+	"cloud.google.com/go/compute/metadata"
```
lookuptable: This should be done by bazel

certmanager/util.go
```diff
@@ -18,10 +18,43 @@ import (
...
+	fetcher.aud = audience
+}
+
+func (fetcher GcpTokenFetcher) setServiceAccount(sa string) {
+	fetcher.serviceAccount = sa
+}
+
+func (fetcher GcpTokenFetcher) getTokenUri() string {
+	// The GCE metadata service URI to get identity token.
+	return "instance/" + fetcher.serviceAccount + "/default/identity?audience=" + fetcher.aud
```
lookuptable: Sorry for raising this again (and let me know if I've misunderstood what you're trying to do).

According to https://cloud.google.com/compute/docs/storing-retrieving-metadata, the URL for a service account named `foo` should be `blah/instance/service-accounts/foo/identity?...`. However, here the uri is constructed as `blah/instance/foo/default/identity?...`. Can you double-check which one is correct?

certmanager/util.go
wattli: done.

certmanager/util.go
wattli: Yes, good catch.

certmanager/util.go
```diff
@@ -19,9 +19,40 @@ import (
...
+	fetcher.aud = audience
+}
+
+func (fetcher *GcpTokenFetcher) setServiceAccount(sa string) {
+	fetcher.serviceAccount = sa
+}
+
+func (fetcher *GcpTokenFetcher) getTokenUri() string {
+	// The GCE metadata service URI to get identity token.
+	return "instance/service-accounts" + fetcher.serviceAccount + "/identity?audience=" + fetcher.aud
```
lookuptable: instance/service-accounts/ (missing back-slash)

wattli: done

wattli:
@lookuptable ,can you re-approve it please? Just fix some lint.

Issue #115, opened 2017-05-17
Create e2e for istio-ca


lookuptable:
We should add basic e2e tests for istio-ca. The tests should cover
- [x] When deployed to a Kubernetes cluster, secrets are correctly created for existing service account
- [x] Creating a new service account leads to the creation of corresponding secret
- [x] Cert in created secret can be successfully verified
- [ ] Istio-ca re-creates secret when one is deleted.

sebastienvas:
Are you thinking of Adding this in istio/istio e2e framework ?

wattli:
For item 1, more specifically, secrets should be overwritten if they already exist

lookuptable:
@sebastienvas I'm thinking of adding it here as part of presubmit. Currently the CA command line stuff is not tested.

sebastienvas:
Awesome

wattli:
@lookuptable , is this issue finished?

lookuptable:
Nope. Hasn't get a chance to work on this.

PR #170, opened 2017-07-08
Integration tests: test secret type and secret data content

lookuptable:
This commit is part of #115.


wattli:
Please fix the presubmit error.

PR #173, opened 2017-07-10
Update getTokenURI()

wattli:


certmanager/util.go
```diff
@@ -33,12 +33,16 @@ type GcpTokenFetcher struct {
 	// aud is the unique URI agreed upon by both the instance and the system verifying the instance's identity.
 	// For more info: https://cloud.google.com/compute/docs/instances/verifying-instance-identity
 	aud            string
-	serviceAccount string
 }
 
 func (fetcher *GcpTokenFetcher) getTokenURI() string {
 	// The GCE metadata service URI to get identity token.
-	return "instance/service-accounts/" + fetcher.serviceAccount + "/identity?audience=" + fetcher.aud
```
wattli: just tested it one more time, instance/service-accounts/default/identity is good enough, which gives you the unique ID corresponding to the service account running on the current VM.

certmanager/util.go
```diff
@@ -33,12 +33,16 @@ type GcpTokenFetcher struct {
...
+	return "instance/service-accounts/default/identity?audience=" + fetcher.aud
+}
+
+func (fetcher *GcpTokenFetcher) getServiceAccount() (string, error) {
```
lookuptable: This is a private method but is not called by anyone. Consider adding it when it's actually needed?

certmanager/util.go
```diff
@@ -33,12 +33,16 @@ type GcpTokenFetcher struct {
 	// aud is the unique URI agreed upon by both the instance and the system verifying the instance's identity.
 	// For more info: https://cloud.google.com/compute/docs/instances/verifying-instance-identity
 	aud            string
-	serviceAccount string
 }
 
 func (fetcher *GcpTokenFetcher) getTokenURI() string {
 	// The GCE metadata service URI to get identity token.
```
lookuptable: Can you update the comment to see the URI is for the **default** service account?

wattli: Done.

certmanager/util.go
wattli: okay.

certmanager/util.go
wattli: This func is deleted.

wattli:
@lookuptable , can you take another look? Just fixed the gofmt error.

certmanager/util.go
wattli: And this is a one line function, do you have any suggestion about the unit test?

PR #183, opened 2017-07-22
Return error instead of logging it

lookuptable:
This allows caller to deal with errors directly and makes testing
easier.

### DIFF
--- a/pkg/pki/crypto.go
+++ b/pkg/pki/crypto.go
@@ -18,48 +18,47 @@ import (
 	"crypto"
 	"crypto/x509"
 	"encoding/pem"
-
-	"github.com/golang/glog"
+	"fmt"
 )
 
 // ParsePemEncodedCertificate constructs a `x509.Certificate` object using the
 // given a PEM-encoded certificate,
-func ParsePemEncodedCertificate(certBytes []byte) *x509.Certificate {
+func ParsePemEncodedCertificate(certBytes []byte) (*x509.Certificate, error) {
 	cb, _ := pem.Decode(certBytes)
 	if cb == nil {
-		glog.Fatalf("Invalid PEM encoding for the certificate: %s", certBytes)
+		return nil, fmt.Errorf("Invalid PEM encoded certificate")
 	}
+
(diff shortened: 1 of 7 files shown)


pkg/pki/crypto_test.go
```diff
@@ -60,76 +88,99 @@ iYh1iBeqIBCSLPt8SrpAWGyOMKLN4bMCWFNOaeBFLG/91K8Yo0swSTAOBgNVHQ8B
...
+	for id, c := range testCases {
+		key, err := ParsePemEncodedKey(c.algo, []byte(c.pem))
+		if c.errMsg != "" {
+			if err == nil {
+				t.Errorf("%s: no error is returned", id)
+			} else if c.errMsg != err.Error() {
+				t.Errorf(`%s: Unexpected error message: want "%s" but got "%s"`, id, c.errMsg, err.Error())
+			}
+		} else {
+			if keyType := reflect.TypeOf(key); keyType != c.keyType {
```
wattli: combine this 
else {
   if 

to "else if" directly?

pkg/pki/crypto.go
```diff
@@ -18,48 +18,47 @@ import (
...
+	"fmt"
 )
 
 // ParsePemEncodedCertificate constructs a `x509.Certificate` object using the
 // given a PEM-encoded certificate,
-func ParsePemEncodedCertificate(certBytes []byte) *x509.Certificate {
+func ParsePemEncodedCertificate(certBytes []byte) (*x509.Certificate, error) {
 	cb, _ := pem.Decode(certBytes)
 	if cb == nil {
-		glog.Fatalf("Invalid PEM encoding for the certificate: %s", certBytes)
```
wattli: Good work, i remember Fatalf is mostly used for main file

lookuptable: yeah that's the objective for this PR.

pkg/pki/crypto_test.go
lookuptable: Done

PR #184, opened 2017-07-24
Fixes linters issues with bazel proto genfiles

sebastienvas:


sebastienvas:
@sarvaniv This should help your PR go in. If this PR fails, you should still be able to add it to your branch used for your PR.

PR #188, opened 2017-07-25
Add a script to check BUILD files are properly configured

lookuptable:
And fix issues with existing BUILD files


lookuptable:
hmm since we are copying generate protobuf go code into /proto directory, running `gazelle` always show some BUILD files needs to be updated. I'm closing this PR while looking for a solution.

PR #192, opened 2017-07-27
Update bazel and rules_go

lookuptable:


lookuptable:
@kyessenov can you also take a look this PR in case I missed anything?

PR #206, opened 2017-08-03
Use rules_docker to build docker image

lookuptable:
Also stop using ubuntu as the base image


lookuptable:
Don't have time to fix the test. Close this for now.

PR #209, opened 2017-08-07
Remove Jenkinsfile

lookuptable:
Go prow go


sebastienvas:
We are still using Jenkins for release, so we ll have to wait a little. Thanks for being proactive on it.

PR #212, opened 2017-08-09
Push docker image to docker.io for release

sebastienvas:


bin/push-docker.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+!/bin/bash
```
lookuptable: Is this change intentional?

bin/push-docker.sh
```diff
@@ -16,34 +16,44 @@ function usage() {
   exit 1
 }
 
+function docker_push() {
+  local im="${1}"
+  if [[ "${im}" =~ ^gcr\.io ]]; then
+    gcloud docker -- push "${im}"
+  else
+    docker push ${im}
```
lookuptable: nit: `${im}` should be doubly quoted.

sebastienvas:
Nice catch.

bin/push-docker.sh
sebastienvas: Done

bin/push-docker.sh
sebastienvas: Done

PR #213, opened 2017-08-09
Add Sign method to handle grpc request

lookuptable:


pkg/server/grpc/server.go
```diff
@@ -0,0 +1,59 @@
...
+	"istio.io/auth/pkg/pki"
+	"istio.io/auth/pkg/pki/ca"
+	pb "istio.io/auth/proto"
+)
+
+type server struct {
+	ca ca.CertificateAuthority
+}
+
+func (s *server) Sign(ctx context.Context, request *pb.Request) (*pb.Response, error) {
```
wattli: This func does actually two things, validation and sign, and you have Sign method in s.ca.Sign as well, which can be confusing. Considering renaming to make it accurate? Like validateAndSign or something?

pkg/server/grpc/server.go
```diff
@@ -0,0 +1,59 @@
...
+	// TODO: handle authenticate here
```
wattli: s/authenticate/authentication

pkg/server/grpc/server.go
lookuptable: The name of the method is defined in protobuf. 

Maybe we can rename it to something generic like `HandleRequest`?

lookuptable: Talked off line: we'll do `HandleCSR()`

PR #214, opened 2017-08-09
Verify certificate fields during integration tests

lookuptable:


integration/main.go
```diff
@@ -190,7 +191,12 @@ func examineSecret(secret *v1.Secret, expectedID string) {
 	key := secret.Data[controller.PrivateKeyID]
 	cert := secret.Data[controller.CertChainID]
 	root := secret.Data[controller.RootCertID]
-	if err := testutil.VerifyCertificate(key, cert, root, expectedID, nil); err != nil {
+	verifyFields := &testutil.VerifyFields{
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
+		KeyUsage:    x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
+		IsCA:        false,
```
wattli: Seems you need to specify the org and timestamp fields.

lookuptable: yeah. I feel we should not do deep equal on the whole object. Instead only compare items present in `VerifyFields`

PR #219, opened 2017-08-14
Git ignore generated protobuf files

lookuptable:
**Release note**:

```
NONE
```


lookuptable:
Merging this without review since it's trivial.

PR #221, opened 2017-08-15
Run GRPC server over TLS

lookuptable:
**Release note**:

```
GRPC service is now served over TLS
```

### DIFF
--- a/pkg/pki/ca/ca.go
+++ b/pkg/pki/ca/ca.go
@@ -36,7 +36,7 @@ const (
 
 // CertificateAuthority contains methods to be supported by a CA.
 type CertificateAuthority interface {
-	Sign(csr *x509.CertificateRequest) ([]byte, error)
+	Sign(csrPEM []byte) ([]byte, error)
 	Generate(name, namespace string) (chain, key []byte)
 	GetRootCertificate() []byte
 }
@@ -137,8 +137,14 @@ func (ca *IstioCA) GetRootCertificate() []byte {
 	return copyBytes(ca.rootCertBytes)
 }
 
-// Sign signs a CSR
-func (ca *IstioCA) Sign(csr *x509.CertificateRequest) ([]byte, error) {
+// Sign takes a PEM-encoded certificate signing request and returns a signed
+// certificate.
+func (ca *IstioCA) Sign(csrPEM []byte) ([]byte, error) {
+	csr, err := pki.ParsePemEncodedCSR(csrPEM)
+	if err != nil {
+		return nil, err
+	}
+
 	tmpl := ca.generateCertificateTemplate(csr)
 
 	bytes, err := x509.CreateCertificate(rand.Reader, tmpl, ca.signingCert, csr.PublicKey, ca.signingKey)
--- a/pkg/server/grpc/server.go
+++ b/pkg/server/grpc/server.go
@@ -31,6 +31,8 @@ import (
 	pb "istio.io/auth/proto"
 )
 
+const certValidationBuffer = time.Minute
+
 // Server implements pb.IstioCAService and provides the service on the
 // specified port.
 type Server struct {
@@ -100,7 +102,7 @@ func (s *Server) createTLSServerOption() grpc.ServerOption {
 	config := &tls.Config{
 		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
 			if s.certificate == nil || !isValid(s.certificate) {
-				// apply new certificate if there isn't one yet, or the one has become invalid
+				// Apply new certificate if there isn't one yet, or the one has become invalid.
 				newCert, err := s.applyServerCertificate()
 				if err != nil {
 					return nil, err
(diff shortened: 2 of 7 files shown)


pkg/pki/ca/ca.go
```diff
@@ -137,8 +137,14 @@ func (ca *IstioCA) GetRootCertificate() []byte {
...
 }
 
-// Sign signs a CSR
-func (ca *IstioCA) Sign(csr *x509.CertificateRequest) ([]byte, error) {
+// Sign takes a PEM-encoded certificate signing request and returns a signed
+// certificate.
+func (ca *IstioCA) Sign(csrPEM []byte) ([]byte, error) {
+	csr, err := pki.ParsePemEncodedCSR(csrPEM)
+	if err != nil {
+		return nil, err
```
wattli: Add some logging info for debugging purpose?

pkg/server/grpc/server.go
```diff
@@ -88,6 +88,59 @@ func (s *Server) Run() error {
...
+		hostname: hostname,
+		port:     port,
+	}
+}
+
+func (s *Server) createTLSServerOption() grpc.ServerOption {
+	config := &tls.Config{
+		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+			if s.certificate == nil || !isValid(s.certificate) {
+				// apply new certificate if there isn't one yet, or the one has become invalid
```
wattli: s/apply/Apply

pkg/server/grpc/server.go
```diff
@@ -88,6 +88,59 @@ func (s *Server) Run() error {
...
+	return grpc.Creds(credentials.NewTLS(config))
+}
+
+func (s *Server) applyServerCertificate() (*tls.Certificate, error) {
+	opts := ca.CertOptions{
+		Host:       s.hostname,
+		RSAKeySize: 1024,
+	}
+
+	csrPEM, privPEM, err := ca.GenCSR(opts)
```
wattli: Is this the self-signed cert option for CA?

pkg/server/grpc/server.go
```diff
@@ -88,6 +88,59 @@ func (s *Server) Run() error {
...
+				}
+				s.certificate = newCert
+			}
+			return s.certificate, nil
+		},
+	}
+	return grpc.Creds(credentials.NewTLS(config))
+}
+
+func (s *Server) applyServerCertificate() (*tls.Certificate, error) {
```
wattli: I think the CA should get key/cert from some file via volume mount, no?

pkg/pki/ca/ca.go
lookuptable: It'll logged by the caller.

pkg/server/grpc/server.go
lookuptable: It's rooted at the same CA that is used for signing workloads, regardless that CA is self-signed or not.

pkg/server/grpc/server.go
```diff
@@ -88,6 +88,59 @@ func (s *Server) Run() error {
...
+	return &cert, nil
+}
+
+func isValid(cert *tls.Certificate) bool {
+	leaf := cert.Leaf
+	if leaf == nil {
+		return false
+	}
+	now := time.Now()
+	return leaf.NotBefore.Before(now) && leaf.NotAfter.After(now)
```
lizan: Leave some buffer time here, to make sure when client receives the certificate it is still valid. (30secs or 60secs should do it).

pkg/server/grpc/server.go
```diff
@@ -88,6 +88,59 @@ func (s *Server) Run() error {
...
+			return s.certificate, nil
+		},
+	}
+	return grpc.Creds(credentials.NewTLS(config))
+}
+
+func (s *Server) applyServerCertificate() (*tls.Certificate, error) {
+	opts := ca.CertOptions{
+		Host:       s.hostname,
+		RSAKeySize: 1024,
```
lizan: Thought 1024-bit RSA is not considered secure enough [(ref)](https://knowledge.rapidssl.com/support/ssl-certificate-support/index?page=content&id=INFO1684&actp=search&viewlocale=en_US&searchid=1442967219430), use 2048?

lookuptable:
@wattli @lizan PTAL

pkg/server/grpc/server.go
```diff
@@ -88,6 +90,60 @@ func (s *Server) Run() error {
...
+	return &cert, nil
+}
+
+func isValid(cert *tls.Certificate) bool {
+	leaf := cert.Leaf
+	if leaf == nil {
+		return false
+	}
+	now := time.Now()
+	return leaf.NotBefore.Add(certValidationBuffer).Before(now) &&
```
lizan: Do we need the buffer for `NotBefore`? Won't this cause a certificate to be considered invalid right after issuance?

lookuptable: I see.

Thinking a little more, I guess it makes sense to rename this method to `needToRefreshCert()`, and skip `NotBefore` check totally.

PR #223, opened 2017-08-17
Remove unnecessary check for certChain parameter since it can be empty

wattli:

### DIFF
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -160,7 +160,7 @@ func createCA() ca.CertificateAuthority {
 
 	certChainBytes := []byte{0}
 	if opts.certChainFile != "" {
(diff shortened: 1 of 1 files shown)


lookuptable:
You'll need to make the file reading optional here: https://github.com/istio/auth/blob/master/cmd/istio_ca/main.go#L162. Otherwise I believe it'll error out.

wattli:
Good catch, updated.

cmd/istio_ca/main.go
```diff
@@ -158,13 +158,18 @@ func createCA() ca.CertificateAuthority {
 		return ca
 	}
 
+	certChainBytes := []byte{0}
```
lookuptable: nit: just do `var certChainBytes []byte`

lookuptable:
\lgtm

PR #225, opened 2017-08-17
Add retries for node agent cert requests. Add grace period to the cert renew interval.

myidpt:
Tests with some necessary refactoring will be added in the next PR.

```release-note
NONE
```

### DIFF
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -160,7 +160,7 @@ func createCA() ca.CertificateAuthority {
 
 	var certChainBytes []byte
 	if opts.certChainFile != "" {
-	        certChainBytes = readFile(opts.certChainFile)
+		certChainBytes = readFile(opts.certChainFile)
 	}
 	caOpts := &ca.IstioCAOptions{
 		CertChainBytes:   certChainBytes,
--- a/node_agent/main.go
+++ b/node_agent/main.go
@@ -37,8 +37,8 @@ func init() {
 		"Istio CA address")
 	naConfig.CertRenewalGracePeriodPercentage = flag.Int("renew-grace-period-percent",
 		na.CertRenewalGracePeriodMaxPercentage,
-		"Integer in [0, 100), grace period length in the percentage of cert TTL. For example, "+
-			"30 means renew when the 70% of the cert's TTL has passed. Default value is 50.")
+		"Integer in [1, 99], grace period length in the percentage of cert TTL. For example, "+
+			"30 means renewing when the 70% of the cert's TTL has passed. Default value is 50.")
 	naConfig.ServiceIdentityDir = flag.String("cert-dir", "./", "Certificate directory")
 	naConfig.RootCACertFile = flag.String("root-cert", "", "Root Certi file")
 	naConfig.Env = flag.Int("env", na.ONPREM, "Node Environment : onprem | gcp")
--- a/node_agent/main.go
+++ b/node_agent/main.go
@@ -35,10 +35,6 @@ func init() {
 		"Node identity private key file")
 	naConfig.IstioCAAddress = flag.String("ca-address", "127.0.0.1",
 		"Istio CA address")
-	naConfig.CertRenewalGracePeriodPercentage = flag.Int("renew-grace-period-percent",
-		na.CertRenewalGracePeriodMaxPercentage,
-		"Integer in [1, 99], grace period length in the percentage of cert TTL. For example, "+
-			"30 means renewing when the 70% of the cert's TTL has passed. Default value is 50.")
 	naConfig.ServiceIdentityDir = flag.String("cert-dir", "./", "Certificate directory")
 	naConfig.RootCACertFile = flag.String("root-cert", "", "Root Certi file")
 	naConfig.Env = flag.Int("env", na.ONPREM, "Node Environment : onprem | gcp")
--- a/node_agent/main.go
+++ b/node_agent/main.go
@@ -35,8 +35,10 @@ func init() {
 		"Node identity private key file")
 	naConfig.IstioCAAddress = flag.String("ca-address", "127.0.0.1",
 		"Istio CA address")
-	naConfig.PercentageExpirationTime = flag.Int("renew-wait-time", 60,
-		"Percentage time to elapse before renewing certificate")
+	naConfig.CertRenewalGracePeriodPercentage = flag.Int("renew-grace-period-percent",
+		na.CertRenewalGracePeriodMaxPercentage,
+		"Integer in [0, 100), grace period length in the percentage of cert TTL. For example, "+
+			"30 means renew when the 70% of the cert's TTL has passed. Default value is 50.")
 	naConfig.ServiceIdentityDir = flag.String("cert-dir", "./", "Certificate directory")
 	naConfig.RootCACertFile = flag.String("root-cert", "", "Root Certi file")
 	naConfig.Env = flag.Int("env", na.ONPREM, "Node Environment : onprem | gcp")
--- a/node_agent/na/nodeagent.go
+++ b/node_agent/na/nodeagent.go
@@ -31,13 +31,13 @@ const (
 	ONPREM int = iota // 0
 	// GCP Node Agent
 	GCP // 1
-	// CertRequestRetrialInterval is the retrial interval for certificate requests.
-	CertRequestRetrialInterval = time.Second
-	// CertRequestMaxRetries is the number of retries for certificate requests.
-	CertRequestMaxRetries = 5
-	// CertRenewalGracePeriodPercentage indicates the length of the grace period in the
+	// certRequestRetrialInterval is the retrial interval for certificate requests.
+	certRequestRetrialInterval = time.Second
+	// certRequestMaxRetries is the number of retries for certificate requests.
+	certRequestMaxRetries = 5
+	// certRenewalGracePeriodPercentage indicates the length of the grace period in the
 	// percentage of the entire certificate TTL.
-	CertRenewalGracePeriodPercentage = 50
+	certRenewalGracePeriodPercentage = 50
 )
 
 // Config is Node agent configuration that is provided from CLI.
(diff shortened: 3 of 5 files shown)


node_agent/na/nodeagent.go
```diff
@@ -92,12 +101,34 @@ func (na *nodeAgentInternal) Start() {
 		glog.Fatalf("Node Agent is not running on the right platform")
 	}
 
-	for {
+	gracePeriodPercentage := *na.config.CertRenewalGracePeriodPercentage
```
wattli: I would worth to add some comments about this logic to make it more readable, thanks.

node_agent/na/nodeagent.go
```diff
@@ -92,12 +101,34 @@ func (na *nodeAgentInternal) Start() {
...
+	if gracePeriodPercentage >= 100 || gracePeriodPercentage < 0 {
+		glog.Errorf("Certificate grace period config %d exceeds limit [0, 99]. Set to default: %d",
```
wattli: The limit is actually [1, 99]

node_agent/na/nodeagent.go
```diff
@@ -92,12 +101,34 @@ func (na *nodeAgentInternal) Start() {
...
 			glog.Errorf("CSR signing failed: %s", err)
-		} else if resp != nil && resp.IsApproved {
-			timer := time.NewTimer(na.getExpTime(resp))
+			retries++
+			timer := time.NewTimer(interval)
+			// Expernentially increase the backoff time.
+			interval = interval * 2
+			<-timer.C
+		}
+		if resp != nil && resp.IsApproved {
```
wattli: It should be a 'else if', right?

node_agent/na/nodeagent.go
```diff
@@ -92,12 +101,34 @@ func (na *nodeAgentInternal) Start() {
...
+			ttl, err := pki.GetPemEncodedCertificateTTL(resp.SignedCertChain)
```
wattli: Please log some info to indicate the CSR is accepted.

wattli: For debugging/testing purpose.

node_agent/na/nodeagent.go
```diff
@@ -92,12 +101,34 @@ func (na *nodeAgentInternal) Start() {
...
+			if err != nil {
+				glog.Fatalf("Error getting TTL from approved cert: %s", err)
```
lookuptable: If ttl is invalid, should we retry instead of quitting the agent completely?

node_agent/main.go
```diff
@@ -35,8 +35,10 @@ func init() {
 		"Node identity private key file")
 	naConfig.IstioCAAddress = flag.String("ca-address", "127.0.0.1",
 		"Istio CA address")
-	naConfig.PercentageExpirationTime = flag.Int("renew-wait-time", 60,
-		"Percentage time to elapse before renewing certificate")
+	naConfig.CertRenewalGracePeriodPercentage = flag.Int("renew-grace-period-percent",
```
lookuptable: I meant picking a default as a constant, instead of making it configurable. Flexibility could lead to confusion and I don't think anyone really wants to configure this value so exposing an option may clog the CLIs.

Anyway I don't feel super strong about this so it's up to you.

node_agent/main.go
```diff
@@ -35,8 +35,10 @@ func init() {
...
+		na.CertRenewalGracePeriodMaxPercentage,
```
lookuptable: Why is this called `Max`Percentage?

node_agent/na/nodeagent.go
```diff
@@ -92,13 +101,48 @@ func (na *nodeAgentInternal) Start() {
...
+		glog.Info("Sending CSR (retrial #%d) ...", retries)
 		privKey, resp, err := na.sendCSR()
-		if err != nil {
-			glog.Errorf("CSR signing failed: %s", err)
-		} else if resp != nil && resp.IsApproved {
-			timer := time.NewTimer(na.getExpTime(resp))
+		if err == nil && resp != nil && resp.IsApproved {
+			cert, certErr := pki.ParsePemEncodedCertificate(resp.SignedCertChain)
+			if certErr != nil {
+				glog.Fatalf("Error getting TTL from approved cert: %s", certErr)
```
lookuptable: This should also be retriable?

cmd/istio_ca/main.go
```diff
@@ -160,7 +160,7 @@ func createCA() ca.CertificateAuthority {
 
 	var certChainBytes []byte
 	if opts.certChainFile != "" {
-	        certChainBytes = readFile(opts.certChainFile)
+		certChainBytes = readFile(opts.certChainFile)
```
wattli: This is fixed, just merge with latest.

node_agent/main.go
```diff
@@ -35,8 +35,10 @@ func init() {
...
+		"Integer in [1, 99], grace period length in the percentage of cert TTL. For example, "+
```
wattli: How about making the logic a little bit simpler, i.e., 30 means renewing when 30% of the TTL has passed? That would make here, and the actual code below easier (i.e., you don't need 100-na.CertRenewalGracePeriodMaxPercentage, but just use na.CertRenewalGracePeriodMaxPercentage directly)

node_agent/main.go
wattli: +1, as a user, i would like to see less flags of a binary.

node_agent/na/nodeagent.go
```diff
@@ -92,13 +101,48 @@ func (na *nodeAgentInternal) Start() {
...
+			waittime := certTTL - time.Duration(gracePeriodPercentage/100)*certTTL
+			timer := time.NewTimer(waittime)
 			na.writeToFile(privKey, resp.SignedCertChain)
+			glog.Info("Obtained new cert. Will renew in %s", waittime.String())
+			retries = 0
+			retrialInterval = CertRequestRetrialInterval
+			<-timer.C
+		} else {
+			if retries >= CertRequestMaxRetries {
+				glog.Fatalf("Max number of retrials has been reached: %d. Abort.", CertRequestMaxRetries)
```
wattli: How about 'Node agent can't get the CSR approved from Istio CA after max number of retrials, please fix the error and retry later.'

node_agent/na/nodeagent.go
```diff
@@ -92,13 +101,48 @@ func (na *nodeAgentInternal) Start() {
...
+			cert, certErr := pki.ParsePemEncodedCertificate(resp.SignedCertChain)
+			if certErr != nil {
+				glog.Fatalf("Error getting TTL from approved cert: %s", certErr)
+			}
+			certTTL := cert.NotAfter.Sub(cert.NotBefore)
+			// Wait until the grace period starts.
+			waittime := certTTL - time.Duration(gracePeriodPercentage/100)*certTTL
+			timer := time.NewTimer(waittime)
 			na.writeToFile(privKey, resp.SignedCertChain)
+			glog.Info("Obtained new cert. Will renew in %s", waittime.String())
```
wattli: Infof?

node_agent/na/nodeagent.go
wattli: But if it is on the server side, it would be confusing to crash the node agent process, right?

node_agent/na/nodeagent.go
```diff
@@ -30,6 +31,13 @@ const (
 	ONPREM int = iota // 0
 	// GCP Node Agent
 	GCP // 1
+	// CertRequestRetrialInterval is the retrial interval for certificate requests.
+	CertRequestRetrialInterval = time.Second
+	// CertRequestMaxRetries is the number of retries for certificate requests.
+	CertRequestMaxRetries = 5
+	// CertRenewalGracePeriodPercentage indicates the length of the grace period in the
+	// percentage of the entire certificate TTL.
+	CertRenewalGracePeriodPercentage = 50
```
lookuptable: Please make these consts private.

node_agent/na/nodeagent.go
```diff
@@ -92,13 +97,41 @@ func (na *nodeAgentInternal) Start() {
...
+			if err != nil {
+				glog.Errorf("CSR signing failed: %s. Will retry in %s", err, retrialInterval.String())
+			} else if resp == nil {
+				glog.Errorf("CSR signing failed: response empty. Will retry in %s", retrialInterval.String())
+			} else {
+				glog.Errorf("CSR signing failed: request not approved. Will retry in %s", retrialInterval.String())
+			}
+			timer := time.NewTimer(retrialInterval)
+			retries++
+			// Expernentially increase the backoff time.
```
lookuptable: s/Expernentially/Exponentially/

node_agent/na/nodeagent.go
```diff
@@ -92,13 +97,41 @@ func (na *nodeAgentInternal) Start() {
...
-		} else if resp != nil && resp.IsApproved {
-			timer := time.NewTimer(na.getExpTime(resp))
+		if err == nil && resp != nil && resp.IsApproved {
+			cert, certErr := pki.ParsePemEncodedCertificate(resp.SignedCertChain)
+			if certErr != nil {
+				glog.Fatalf("Error getting TTL from approved cert: %s", certErr)
+			}
+			certTTL := cert.NotAfter.Sub(cert.NotBefore)
+			// Wait until the grace period starts.
+			waittime := certTTL - time.Duration(CertRenewalGracePeriodPercentage/100)*certTTL
```
lookuptable: nit: waitTime (camel case for consistency?)

node_agent/na/nodeagent.go
myidpt: Done.

node_agent/na/nodeagent.go
myidpt: Done.

node_agent/na/nodeagent.go
myidpt: Done.

node_agent/na/nodeagent.go
wattli: Please address, thanks.

myidpt: The client side crashes anyway if the server has internal error, it's either immediately or after 5 retires. I can put retry here. IMO, this is relatively rare case, and it's not a big deal whether we do retry or not.

wattli: sg

node_agent/na/nodeagent.go
wattli: please fix

node_agent/na/nodeagent.go
```diff
@@ -92,15 +97,52 @@ func (na *nodeAgentInternal) Start() {
...
+			success = false
+		}
+
+		if !success {
+			if retries >= certRequestMaxRetries {
+				glog.Fatalf("Node agent can't get the CSR approved from Istio CA after max number of retrials"+
+					"(%d), please fix the error and retry later.", certRequestMaxRetries)
+			}
+			if err != nil {
+				glog.Errorf("CSR signing failed: %s. Will retry in %s", err, retrialInterval.String())
```
lookuptable: nit: error should be printed as %v, not %s

myidpt: Done.

PR #229, opened 2017-08-21
Add a method to extract identities from SAN extension

lookuptable:
This will be used by the grpc server for client certificate based
authN/authZ.

Also grouping all SAN-related operations into package `pkg/pki/san.go`

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/pki/san.go
+++ b/pkg/pki/san.go
@@ -1,0 +1,153 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pki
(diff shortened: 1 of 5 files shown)


pkg/pki/san.go
```diff
@@ -0,0 +1,153 @@
...
+func ExtractSANExtension(cert *x509.Certificate) (*pkix.Extension, bool) {
+	for _, ext := range cert.Extensions {
+		if ext.Id.Equal(oidSubjectAlternativeName) {
+			// We don't need to examine other extensions anymore since a certificate
+			// must not include more than one instance of a particular extension. See
+			// https://tools.ietf.org/html/rfc5280#section-4.2.
+			return &ext, true
+		}
+	}
+	return nil, false
```
wattli: I feel the bool is unnecessary, nil means false already, right?

lookuptable: Yeah. we can check for nil as well.

yutongz:
@wattli  Could you comment them in two different lines

yutongz:
The pr got merged, yeah!!  Thank you guys for trusting the merge robot.
So basically, we you want to actually merge the pr, you need to comment "/lgtm" and "approve"
The reason why we need two level approvals is to bring more flexible. For example, you can ask Wencheng and Tao to review this at the same time. Wencheng can say "lgtm" after he is ok with it. But Tao may have some nit changes. I understand in many case, one pr only has one reviewer, especially for auth repo. So if you really don't like the two level approval, we can totally change to only "/lgtm" is required.
Feel free to bug me when you have problems with this robot.

PR #230, opened 2017-08-21
Add a GRPC authenticator/authorizer for client certificate

lookuptable:
**Release note**:

```
Add an client certificate based authentication and authorization for the GRPC server.
```


pkg/pki/san.go
```diff
@@ -131,9 +130,9 @@ func ExtractIDsFromSAN(sanExt *pkix.Extension) ([]Identity, error) {
 }
 
 // ExtractSANExtension extracts the "Subject Alternative Name" externsion from
-// the given X.509 cerificate.
-func ExtractSANExtension(cert *x509.Certificate) *pkix.Extension {
-	for _, ext := range cert.Extensions {
+// the given extension set.
```
wattli: Use "the given pki extension set"?

pkg/server/grpc/authorizer.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (authZ *simpleAuthorizer) authorize(requester *user, requestedIDs []string) bool {
+	idMap := make(map[string]bool, len(requester.identities))
+	for _, id := range requester.identities {
+		idMap[id] = true
+	}
+
+	for _, requestedID := range requestedIDs {
+		if _, exists := idMap[requestedID]; !exists {
+			glog.Infof("The requested identity (%q) does not belong to the requester", requestedID)
```
wattli: What about "The requested identity does not *match* the requester?"

And what about using glog.Warningf ?

pkg/server/grpc/server.go
```diff
@@ -92,14 +117,21 @@ func (s *Server) Run() error {
...
 	}
 }
 
 func (s *Server) createTLSServerOption() grpc.ServerOption {
+	cp := x509.NewCertPool()
+	cp.AppendCertsFromPEM(s.ca.GetRootCertificate())
+
 	config := &tls.Config{
+		ClientCAs:  cp,
+		ClientAuth: tls.VerifyClientCertIfGiven,
```
wattli: This logic needs to be carefully handled (i.e., for on prem, we will check the client cert must be presented, and for gcp, the jwt must be presented). Does not have to be in this PR, but just a reminder for both of us :)

lookuptable: I believe we should set `VerifyClientCertIfGiven` here instead of setting `AlwasyRequireClientCert`.

We have an authenticator on top of tls that actually makes use of the verified certificate. So for on prem, the authenticator will reject the request in case a client cert is missing.

pkg/server/grpc/util.go
```diff
@@ -0,0 +1,46 @@
...
+	idsWithType, err := pki.ExtractIDsFromSAN(sanExt)
+	if err != nil {
+		glog.Warningf("failed to extract identities from SAN extension (error %v)", err)
+
+		return nil
+	}
+
+	ids := []string{}
+	for _, id := range idsWithType {
+		ids = append(ids, string(id.Value))
```
myidpt: Maybe you want to verify the id type is URI before appending?

lookuptable: I feel we don't care the type here. The type is mainly used to dictate how to encode id into cert. But here we only care the id string.

lookuptable:
@wattli PTAL

pkg/server/grpc/authorizer.go
wattli: ping

PR #235, opened 2017-08-22
Add config to build Auth release on GCB

rvkubiak:
**Release note**:
```release-note
Add a config file for running Auth release builds on Google Container Builder.
```


bin/build_auth_release.yaml
```diff
@@ -0,0 +1,8 @@
+# Used for automated builds triggered on Git tags pushed matching the regex:
+#   \d+\.\d+\.\d+
+
+steps:
+- name: 'gcr.io/$PROJECT_ID/istio-builder'
+  args: [ 'bash', './bin/push-docker.sh', '-h', "gcr.io/$PROJECT_ID", '-t', "$TAG_NAME" ]
```
sebastienvas: It would be good to create a release.sh, but if not use gcr.io/istio-io,docker.io/istio for -h

rvkubiak: Hard coded the project name.  Will add docker.io once the credential work is done.

rvkubiak: Added release.sh script and DockerHub push.

sebastienvas:
/lgtm 
Need one more approver
/assign @lookuptable

PR #237, opened 2017-08-22
Return intermediate certs when CA signs a CSR

lookuptable:
**Release note**:

```
NONE
```
/assign @wattli


pkg/pki/ca/ca.go
```diff
@@ -156,7 +156,12 @@ func (ca *IstioCA) Sign(csrPEM []byte) ([]byte, error) {
 		Type:  "CERTIFICATE",
 		Bytes: bytes,
 	}
-	return pem.EncodeToMemory(block), nil
+	cert := pem.EncodeToMemory(block)
+
+	// Also append intermediate certs into the chain.
+	chain := append(cert, ca.certChainBytes...)
```
wattli: The order actually matters, it should be cert-chain first, and then leaf.

wattli: And this is not enough, we should also append ca's signing cert (no root), of course, on the top.

And since you are already here, can you help to fix line 130 in this file as well? Thanks

wattli: Turns out it's some misunderstanding, will have refactoring on this.

lookuptable:
@wattli can you sign off this?

PR #243, opened 2017-08-24
Refactor NodeAgent and add unit tests

myidpt:
**Release note**:
```release-note
Refactor NodeAgent and add unit tests.
```


cmd/node_agent/na/nodeagent.go
```diff
@@ -78,43 +79,98 @@ type platformSpecificRequest interface {
 	IsProperPlatform() bool
 }
 
+// CAGrpcClient is for implementing the GRPC client to talk to CA.
+type CAGrpcClient interface {
```
lookuptable: Why do we need this here? Why not use the one generated from protobuf?

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,43 +79,98 @@ type platformSpecificRequest interface {
...
+	// Send CSR to the CA and gets the response or error.
+	SendCSR(*string, []grpc.DialOption, *pb.Request) (*pb.Response, error)
+}
+
+// CAGrpcClientImpl is a implementation of GRPC client to talk to CA.
+type CAGrpcClientImpl struct {
```
lookuptable: nit: make it private?

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,43 +79,98 @@ type platformSpecificRequest interface {
...
 
-// Start the node Agent.
-func (na *nodeAgentInternal) Start() {
+// Start the node Agent with default setups.
+func (na *nodeAgentInternal) Start() error {
+	return na.StartWithArgs(certRequestRetrialInterval, certRequestMaxRetries, certRenewalGracePeriodPercentage)
+}
 
+// Start the node Agent with configs about retries.
+func (na *nodeAgentInternal) StartWithArgs(interval time.Duration, maxRetries int, gracePeriodPercentage int) error {
```
lookuptable: Can we make these parameters members of the struct? So you'll do
```
na := &nodeAgentInternal{
  interval: ...
  maxRetries: ...
  gracePeriodPercentage...
}
na.Start()
```

lookuptable: Also do these parameters belong to `na.config`?

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,43 +79,98 @@ type platformSpecificRequest interface {
...
+
+// CAGrpcClientImpl is a implementation of GRPC client to talk to CA.
+type CAGrpcClientImpl struct {
+}
+
+// SendCSR sends CSR to CA through GRPC.
+func (c *CAGrpcClientImpl) SendCSR(address *string, options []grpc.DialOption, req *pb.Request) (*pb.Response, error) {
+	conn, err := grpc.Dial(*address, options...)
+	if err != nil {
+		glog.Errorf("Failed to dial %s: %s", *address, err)
```
lookuptable: Guess you don't need both return and log the error. Otherwise an error will be logged twice since everything is captured here:https://github.com/istio/auth/pull/243/files#diff-45d9bd931e75161931bda07cf3f93c48R60

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,43 +79,98 @@ type platformSpecificRequest interface {
...
+	// Send CSR to the CA and gets the response or error.
+	SendCSR(*string, []grpc.DialOption, *pb.Request) (*pb.Response, error)
+}
+
+// CAGrpcClientImpl is a implementation of GRPC client to talk to CA.
+type CAGrpcClientImpl struct {
+}
+
+// SendCSR sends CSR to CA through GRPC.
+func (c *CAGrpcClientImpl) SendCSR(address *string, options []grpc.DialOption, req *pb.Request) (*pb.Response, error) {
```
lookuptable: Are we expecting the address and options to change? If not shall we set them as members of `CAGrpcClientImpl` struct?

cmd/node_agent/na/nodeagent.go
```diff
@@ -148,57 +206,47 @@ func (na *nodeAgentInternal) Start() {
...
+	if waitTime < 0 {
+		waitTime = 0
 	}
-
-	return privKey, resp, nil
+	return waitTime, nil
 }
 
-func (na *nodeAgentInternal) writeToFile(privKey []byte, cert []byte) {
+func (na *nodeAgentInternal) writeToFile(privKey []byte, cert []byte) error {
```
lookuptable: nit: this doesn't need to be a member function?

cmd/node_agent/na/nodeagent_test.go
```diff
@@ -0,0 +1,208 @@
...
+			cAClient:    &FakeCAClient{0, &pb.Response{IsApproved: true, SignedCertChain: []byte(`INVALIDCERT`)}, nil},
+			expectedErr: "node agent can't get the CSR approved from Istio CA after max number of retries (3)",
+			sendTimes:   4,
+		},
+	}
+
+	for id, c := range testCases {
+		na := nodeAgentInternal{c.config, c.req, c.cAClient}
+		err := na.StartWithArgs(time.Millisecond, 3, 50)
+		if strings.Compare(err.Error(), c.expectedErr) != 0 {
```
lookuptable: You can just use `!=` here

cmd/node_agent/na/nodeagent.go
myidpt: I have moved them to na.config and put Config to a separate file.

cmd/node_agent/na/nodeagent.go
myidpt: I personally think we should have all the file I/O functions in a util/ folder. This reduces duplication of file I/O functions and facilitates our tests. I can refactor it if you think necessary. I think for now, we can temporarily keep it here.

cmd/node_agent/na/nodeagent_test.go
myidpt: Done.

cmd/node_agent/na/nodeagent.go
lookuptable: Sorry I was wrong. The generated one has the following signature: `HandleCSR(ctx context.Context, in *Request, opts ...grpc.CallOption) (*Response, error)`. Guess we need a wrapper here since we're not passing in ctx and opts.

cmd/node_agent/main.go
```diff
@@ -60,5 +62,7 @@ func main() {
 func runNodeAgent() {
 	nodeAgent := na.NewNodeAgent(&naConfig)
 	glog.Infof("Starting Node Agent")
-	nodeAgent.Start()
+	if err := nodeAgent.Start(); err != nil {
+		glog.Errorf("Node agent termitated with error: %v.", err)
```
lookuptable: s/termitated/terminated/

cmd/node_agent/na/nodeagent.go
```diff
@@ -148,57 +172,47 @@ func (na *nodeAgentInternal) Start() {
...
-	}()
-
-	client := pb.NewIstioCAServiceClient(conn)
-	privKey, req := na.createRequest()
-	resp, err := client.HandleCSR(context.Background(), req)
-	if err != nil {
-		glog.Errorf("CSR request failed %v", err)
-		return nil, nil, err
+	gracePeriod := cert.NotAfter.Sub(cert.NotBefore) * time.Duration(gracePeriodPercentage) / time.Duration(100)
+	// Wait until the grace period starts.
```
lookuptable: Is this comment accurate?

cmd/node_agent/na/config.go
```diff
@@ -0,0 +1,75 @@
...
+
+	RSAKeySize int
+
+	// Istio CA grpc server
+	IstioCAAddress string
+
+	// The environment this node agent is running on
+	Env int
+
+	// CSRInitialRetrialInterval is the retrial interval for certificate requests.
```
lookuptable: Sorry I might have confused you but as of now we don't need to put those CSR* variables. We can just directly use the consts defined around line 22.

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,46 +42,106 @@ type platformSpecificRequest interface {
...
+	if optionErr != nil {
+		return nil, optionErr
+	}
+	return &cAGrpcClientImpl{
+		&cfg.IstioCAAddress,
+		dialOptions,
+	}, nil
+}
+
+// cAGrpcClientImpl is a implementation of GRPC client to talk to CA.
```
lookuptable: is an implementation

cmd/node_agent/na/nodeagent.go
```diff
@@ -78,46 +42,106 @@ type platformSpecificRequest interface {
...
+			waitTime, ttlErr := na.getWaitTimeFromCert(
+				resp.SignedCertChain, time.Now(), na.config.CSRGracePeriodPercentage)
+			if ttlErr != nil {
+				glog.Errorf("Error getting TTL from approved cert: %v", ttlErr)
 				success = false
 			} else {
-				certTTL := cert.NotAfter.Sub(cert.NotBefore)
-				// Wait until the grace period starts.
-				waitTime := certTTL - time.Duration(certRenewalGracePeriodPercentage/100)*certTTL
 				timer := time.NewTimer(waitTime)
```
lookuptable: nit: move this closer to where it's used (line 145)

cmd/node_agent/na/nodeagent.go
```diff
@@ -148,57 +172,47 @@ func (na *nodeAgentInternal) Start() {
...
-	privKey, req := na.createRequest()
-	resp, err := client.HandleCSR(context.Background(), req)
-	if err != nil {
-		glog.Errorf("CSR request failed %v", err)
-		return nil, nil, err
+	gracePeriod := cert.NotAfter.Sub(cert.NotBefore) * time.Duration(gracePeriodPercentage) / time.Duration(100)
+	// Wait until the grace period starts.
+	waitTime := timeToExpire - gracePeriod
+	if waitTime < 0 {
+		waitTime = 0
```
lookuptable: We should error out here, no? waitTime less than 0 means we're getting a cert that needs to be refreshed immediately.

cmd/node_agent/main.go
myidpt: Done.

cmd/node_agent/na/nodeagent.go
myidpt: Done.

cmd/node_agent/na/nodeagent.go
myidpt: Rephrased.

cmd/node_agent/na/nodeagent.go
myidpt: Sure, we can return error here.

PR #244, opened 2017-08-24
Unifying CA signing interface

lookuptable:
Currently the CA exposes two methods:
* `Sign(csrPEM []string)` which is used by the GRPC server
* `Generate(name, namespace []string)` which is used for generating
cert based on K8S service account.

However these two methods are very similar, and this PR removes the
second method completely. After this PR, the K8S secret controller
constructs a CSR locally and uses the `Sign` method to get a
certificate.

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/pki/ca/controller/secret.go
+++ b/pkg/pki/ca/controller/secret.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"bytes"
+	"fmt"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/pki/ca/controller/secret.go
```diff
@@ -214,6 +224,26 @@ func (sc *SecretController) scrtDeleted(obj interface{}) {
...
+		Host:       id,
+		RSAKeySize: keySize,
+	}
+
+	csrPEM, keyPEM, err := ca.GenCSR(options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	certPEM, err := sc.ca.Sign(csrPEM)
```
lookuptable: I share the same feeling and I can refactor in a followup PR. 

One proposal is to move `GenCSR` and `GenCert` into a separate `x509` package. Then we'll have `sc.ca.Sign` and `x509.GenCSR`. Do you think this is less confusing?

wattli: sgtm.

PR #248, opened 2017-08-29
Setup proper path for node_agent

wattli:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/node_agent/main.go
+++ b/cmd/node_agent/main.go
@@ -39,14 +39,14 @@ func init() {
 
 	flags := rootCmd.Flags()
 
-	flags.StringVar(&naConfig.ServiceIdentity, "service-identity", "", "Service Identity the node agent is managing")
-	flags.StringVar(&naConfig.ServiceIdentityOrg, "org", "Juju org", "Organization for the cert")
+	flags.StringVar(&naConfig.ServiceIdentity, "service-identity", "default", "Service Identity the node agent is managing")
+	flags.StringVar(&naConfig.ServiceIdentityOrg, "org", "", "Organization for the cert")
 	flags.IntVar(&naConfig.RSAKeySize, "key-size", 1024, "Size of generated private key")
-	flags.StringVar(&naConfig.NodeIdentityCertFile, "na-cert", "", "Node Agent identity cert file")
-	flags.StringVar(&naConfig.NodeIdentityPrivateKeyFile, "na-key", "", "Node identity private key file")
+	flags.StringVar(&naConfig.NodeIdentityCertFile, "na-cert", "/etc/certs/na-cert.pem", "Node Agent identity cert file")
+	flags.StringVar(&naConfig.NodeIdentityPrivateKeyFile, "na-key", "/etc/certs/na-key.pem", "Node identity private key file")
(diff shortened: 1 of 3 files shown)


cmd/node_agent/main.go
```diff
@@ -39,14 +39,14 @@ func init() {
 
 	flags := rootCmd.Flags()
 
-	flags.StringVar(&naConfig.ServiceIdentity, "service-identity", "", "Service Identity the node agent is managing")
-	flags.StringVar(&naConfig.ServiceIdentityOrg, "org", "Juju org", "Organization for the cert")
+	flags.StringVar(&naConfig.ServiceIdentity, "service-identity", "default", "Service Identity the node agent is managing")
```
lookuptable: Is this a reasonable default? I don't issue any identity called "default".

wattli: in the beginning I thought it's the right one, updated.

cmd/node_agent/main.go
```diff
@@ -39,14 +39,20 @@ func init() {
...
-	flags.StringVar(&naConfig.NodeIdentityCertFile, "na-cert", "", "Node Agent identity cert file")
-	flags.StringVar(&naConfig.NodeIdentityPrivateKeyFile, "na-key", "", "Node identity private key file")
-	flags.StringVar(&naConfig.IstioCAAddress, "ca-address", "127.0.0.1", "Istio CA address")
+	flags.StringVar(&naConfig.NodeIdentityCertFile, "na-cert",
+		"/etc/certs/na-cert.pem", "Node Agent identity cert file")
+	flags.StringVar(&naConfig.NodeIdentityPrivateKeyFile,
+		"na-key", "/etc/certs/na-key.pem", "Node identity private key file")
+	flags.StringVar(&naConfig.IstioCAAddress,
+		"ca-address", "istio-ca.default.svc.cluster.local:8060", "Istio CA address")
 	flags.StringVar(&naConfig.ServiceIdentityDir, "cert-dir", "./", "Certificate directory")
```
myidpt: Default /etc/certs/ ?

cmd/node_agent/na/nodeagent.go
```diff
@@ -204,10 +204,10 @@ func (na *nodeAgentInternal) getWaitTimeFromCert(
 
 func (na *nodeAgentInternal) writeToFile(privKey []byte, cert []byte) error {
 	glog.Infof("Write key and cert to local file.")
-	if err := ioutil.WriteFile("serviceIdentityKey.pem", privKey, 0600); err != nil {
+	if err := ioutil.WriteFile("/etc/certs/key.pem", privKey, 0600); err != nil {
```
myidpt: Concatenate ServiceIdentityDir with "Key.pem"?

cmd/node_agent/main.go
wattli: I will remove this flag, we don't need two flags for one file, and that's less readable too.

cmd/node_agent/na/nodeagent.go
wattli: As above

cmd/node_agent/main.go
myidpt: Sure.

PR #255, opened 2017-08-31
Add JWT authentication

lookuptable:
**Release note**:

```
Support node agent authentication using Google ID token.
```


cmd/node_agent/na/gcp_test.go
```diff
@@ -36,10 +36,11 @@ func (fetcher *mockTokenFetcher) FetchToken() (string, error) {
 
 func TestGetDialOptions(t *testing.T) {
 	gcp := gcpPlatformImpl{&mockTokenFetcher{}}
+	config := &Config{RootCACertFile: "testdata/cert-chain-good.pem"}
+	options, _ := gcp.GetDialOptions(config)
```
wattli: also check err is nill?

pkg/server/grpc/authenticator.go
```diff
@@ -15,23 +15,43 @@
...
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
 
 	"istio.io/auth/pkg/pki"
 )
 
+const (
+	httpAuthHeader    = "authorization"
```
wattli: sort

pkg/server/grpc/authenticator.go
```diff
@@ -58,5 +78,68 @@ func (cca *clientCertAuthenticator) authenticate(ctx context.Context) *user {
...
+
+	authHeader, exists := md[httpAuthHeader]
+	if !exists {
+		glog.Warning("no HTTP authorization header exists")
+		return ""
+	}
+
+	for _, value := range authHeader {
+		if strings.HasPrefix(value, bearerTokenPrefix) {
+			return strings.TrimPrefix(value, bearerTokenPrefix)
```
wattli: Is it possible to have two bearer tokens?

lookuptable: I feel this is rarely used. I'm looking into RFCs to see whether it's even allowed.

wattli: Please ping back this thread once you are done.

lookuptable: Dig more about this - though I can't find any RFC that forbids multiple authorization headers, the current grpc-go doesn't support multiple. See here: https://github.com/grpc/grpc-go/blob/master/transport/http2_client.go#L346

cmd/node_agent/na/gcp.go
```diff
@@ -15,20 +15,28 @@
...
+const (
+	bearerTokenScheme = "Bearer"
+	httpAuthHeader    = "authorization"
+)
+
 type jwtAccess struct {
 	token string
 }
 
 func (j *jwtAccess) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
```
myidpt: Where is this used?

pkg/server/grpc/authenticator.go
wattli: sg

pkg/server/grpc/server.go
```diff
@@ -171,6 +180,15 @@ func (s *Server) applyServerCertificate() (*tls.Certificate, error) {
 	return &cert, nil
 }
 
+func (s *Server) authenticate(ctx context.Context) *user {
+	for _, authn := range s.authenticators {
+		if u := authn.authenticate(ctx); u != nil {
```
myidpt: I think you may want to be explicit about the order here. E.g., use client cert authentication if both cert and jwt are presented.

cmd/node_agent/na/gcp.go
lookuptable: This is used by gRPC internally when making a connection.

pkg/server/grpc/server.go
lookuptable: Where/how do you want me to make it explicit? Adding comments to where the authenticators get initialized?

cmd/node_agent/na/gcp.go
myidpt: I see. Thanks.

pkg/server/grpc/server.go
myidpt: Maybe make authenticators a map[int]authenticator ? The int index can be the authenticator type enum. And here you can do:
if s.authenticators[CertAuthenticator] != nil {
  s.authenticators[CertAuthenticator].authenticate(ctx)
  ...
}
if s.authenticators[JwtAuthenticator] != nil {
  s.authenticators[JwtAuthenticator].authenticate(ctx)
  ...
}

lookuptable: hmm I feel this is a verbose way to implement the same logic. if you're worrying about confusing users, I believe clear documentation makes more sense than code change.

myidpt: Sure, doc/comment is definitely helpful. We can use the order of s.authenticators to specify the authenticator usage order. So now it is first cert then JWT, as can be seen from the New() func.

PR #256, opened 2017-09-01
Add comments about the order of authenticators

lookuptable:
**Release note**:

```
NONE
```
/assign @myidpt


myidpt:
Thanks for the quick response!

PR #258, opened 2017-09-01
Rename istio.service to istio-node-agent.service to avoid conflict

wattli:
**Release note**:

```release-note
NONE
```


tools/deb/BUILD
```diff
@@ -11,7 +11,7 @@ pkg_tar(
 
 pkg_tar(
     name = "istio-systemd",
-    files = ["istio.service"],
+    files = ["istio-node-agent.service"],
```
lookuptable: Should we call it `istio-auth-agent`, which sounds more on par with the pilot agent?

wattli: Makes sense. Done

wattli: Can you /lgtm it? Thanks.

tools/deb/istio-auth-node-agent.service
```diff
@@ -1,9 +1,9 @@
 [Unit]
-Description=istio-node-agent: The Istio node agent
+Description=istio-auth-node-agent: The Istio auth node agent
 Documentation=http://istio.io/
 
 [Service]
-ExecStart=/usr/local/bin/istio-node-agent --env=0 --alsologtostder
+ExecStart=/usr/local/bin/istio-auth-node-agent
```
lookuptable: why are the CMD options being removed? Are they specified somewhere else?

wattli: 'env' is unnecessary since we are using the default value, --alsologtostderr is mostly for debugging purpose.

costinm:
Where does the log go by default ?

In systemd, AFAIK it is expected all logs go to stderr - and nowhere else (
journald takes care if saving)

lookuptable:
Can you address Costin's comment before merging?

wattli:
Comment addressed

wattli:
@lookuptable , PTAL

PR #261, opened 2017-09-11
Build and push Debian package as part of the release

rvkubiak:
**Release note**:
```
Build and push Debian package as part of the release
```

### DIFF
--- a/bin/push-debian.sh
+++ b/bin/push-debian.sh
@@ -5,11 +5,12 @@
 # bin/push-debian.sh \
 #   -c opt
 #   -v 0.2.1
+#   -p gs://istio-release/release/0.2.1/deb
 
 function usage() {
   echo "$0 \
     -c <bazel config to use> \
-    -p <bucket path, e.g. gs://istio-release/release/deb> \
+    -p <GCS path, e.g. gs://istio-release/release/0.2.1/deb> \
     -v <istio version number>"
   exit 1
 }
--- a/bin/push-debian.sh
+++ b/bin/push-debian.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Example usage:
+#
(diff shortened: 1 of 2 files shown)


bin/push-debian.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ -z "${BAZEL_ARGS}" ] || [ -z "${ISTIO_VERSION}" ]; then
+  usage
+fi
+
+set -ex
+
+bazel ${BAZEL_STARTUP_ARGS} build ${BAZEL_ARGS} "//tools/deb:istio-auth-node-agent"
+gsutil -m cp -r \
+  bazel-bin/tools/deb/istio-auth-node-agent_${ISTIO_VERSION}_amd64.* \
+  gs://istio-release/releases/${ISTIO_VERSION}/deb
```
sebastienvas: Could you please make configurable like in here:
https://github.com/istio/pilot/blob/master/bin/upload-istioctl#L23

rvkubiak: How do you want this to work?  In upload-istioctl you pass it the full GCS path, however for the debian package the version number is included in the path.  We can do the path prefix, so something like:
push-debian.sh -p gs://istio-release/releases
gsutil -m cp -r *.deb ${BUCKET_PATH}/${ISTIO_VERSION}/deb

I don't really see this as an improvement.  I could see doing something like:
push-debian.sh -b istio-release
gsutil -m cp -r *.deb gs://${BUCKET_NAME}/releases/${ISTIO_VERSION}/deb

Alternatively we could just move all the logic to the release script with something like:
push-debian.sh -b gs://istio-release/releases/0.2.1/deb

rvkubiak: Resolved.

PR #264, opened 2017-09-14
Upload Debian for pre and post submit

sebastienvas:
Update script to use the right tag when creating the debian package.

```release-note
none
```


sebastienvas:
@costinm @ldemailly I am planning to do the same for proxy and pilot to ease release creation.

PR #266, opened 2017-09-19
Store self-signed key/cert in secret so next time CA does not regenerate

wattli:
… write it to secret for persistency

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -56,6 +56,7 @@ type cliOptions struct {
 
 	kubeConfigFile string
 
+	selfSignedCA        bool
 	selfSignedCAOrg string
 
 	caCertTTL time.Duration
@@ -63,9 +64,6 @@ type cliOptions struct {
 
 	grpcHostname string
 	grpcPort     int
-
-	selfSignedCA        bool
-	targetAllNamespaces bool
 }
 
 var (
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -89,8 +89,8 @@ func init() {
 	flags.StringVar(&opts.namespace, "namespace", "",
 		"Select a namespace for the CA to listen to. If unspecified, Istio CA tries to use the ${"+namespaceKey+"} "+
 			"environment variable. If neither is set, Istio CA listens to all namespaces.")
-	flags.StringVar(&opts.istioCaStorageNamespace, "istio-ca-storage-namespace", "istio-system", "Namespace where " +
-			"the Istio CA pods is running. Will not be used if explicit file or other storage mechanism is specified.")
+	flags.StringVar(&opts.istioCaStorageNamespace, "istio-ca-storage-namespace", "istio-system", "Namespace where "+
+		"the Istio CA pods is running. Will not be used if explicit file or other storage mechanism is specified.")
 
 	flags.StringVar(&opts.kubeConfigFile, "kube-config", "",
 		"Specifies path to kubeconfig file. This must be specified when not running inside a Kubernetes pod.")
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -56,7 +56,7 @@ type cliOptions struct {
 
 	kubeConfigFile string
 
-	selfSignedCA        bool
+	selfSignedCA    bool
 	selfSignedCAOrg string
 
 	caCertTTL time.Duration
@@ -169,7 +169,7 @@ func createCA(core corev1.SecretsGetter) ca.CertificateAuthority {
 		ca, err := ca.NewSelfSignedIstioCA(opts.caCertTTL, opts.certTTL, opts.selfSignedCAOrg,
 			opts.caNamespace, core)
 		if err != nil {
-			glog.Fatalf("Failed to create a self-signed Istio CA (error: %v).", err)
+			glog.Fatalf("Failed to create a self-signed Istio CA (error: %v)", err)
 		}
 		return ca
 	}
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -56,9 +56,9 @@ type cliOptions struct {
 
 	selfSignedCA    bool
 	selfSignedCAOrg string
-	
-	caCertTTL       time.Duration
-	certTTL         time.Duration
+
+	caCertTTL time.Duration
+	certTTL   time.Duration
 
 	grpcHostname string
 	grpcPort     int
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -50,9 +50,9 @@ type cliOptions struct {
 	signingKeyFile  string
 	rootCertFile    string
 
-	namespace   string
+	namespace          string
 	targetAllNamespace bool
-	kubeConfigFile string
+	kubeConfigFile     string
 
 	selfSignedCA bool
 
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -56,6 +56,7 @@ type cliOptions struct {
 
 	selfSignedCA    bool
 	selfSignedCAOrg string
+	
 	caCertTTL       time.Duration
 	certTTL         time.Duration
 
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -85,11 +85,12 @@ func init() {
 	flags.StringVar(&opts.rootCertFile, "root-cert", "", "Specifies path to the root certificate file")
 
 	flags.StringVar(&opts.namespace, "namespace", "", "The namesapce that CA is running at.")
-	flags.BoolVar(&opts.targetAllNamespaces, "all-namespaces", false,
-		"Indicates whether to listen to all namespaces or just the namespace CA is running at")
 
 	flags.StringVar(&opts.kubeConfigFile, "kube-config", "",
 		"Specifies path to kubeconfig file. This must be specified when not running inside a Kubernetes pod.")
+
+	flags.BoolVar(&opts.targetAllNamespaces, "all-namespaces", false,
+		"Indicates whether to listen to all namespaces or just the namespace CA is running at")
 
 	flags.BoolVar(&opts.selfSignedCA, "self-signed-ca", false,
 		"Indicates whether to use auto-generated self-signed CA certificate. "+
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -160,7 +160,7 @@ func createCA(core corev1.CoreV1Interface) ca.CertificateAuthority {
 		glog.Info("Use self-signed certificate as the CA certificate")
 
 		// TODO(wattli): Refactor this and combine it with NewIstioCA().
-		ca, err := ca.NewSelfSignedIstioCA(opts.caCertTTL, opts.certTTL, opts.cAOrg, opts.namespace, core)
+		ca, err := ca.NewSelfSignedIstioCA(opts.caCertTTL, opts.certTTL, opts.selfSignedCAOrg, opts.namespace, core)
 		if err != nil {
 			glog.Fatalf("Failed to create a self-signed Istio CA (error: %v)", err)
 		}
@@ -176,7 +176,7 @@ func createCA(core corev1.CoreV1Interface) ca.CertificateAuthority {
 		CertTTL:          opts.certTTL,
 		Core:             core,
 		Namespace:        opts.namespace,
-		Org:              opts.cAOrg,
+		Org:              opts.selfSignedCAOrg,
 		SigningCertBytes: readFile(opts.signingCertFile),
 		SigningKeyBytes:  readFile(opts.signingKeyFile),
 		RootCertBytes:    readFile(opts.rootCertFile),
--- a/cmd/istio_ca/main.go
+++ b/cmd/istio_ca/main.go
@@ -50,9 +50,9 @@ type cliOptions struct {
 	signingKeyFile  string
 	rootCertFile    string
 
-	namespace          string
-	targetAllNamespace bool
-	kubeConfigFile     string
+	namespace           string
+	targetAllNamespaces bool
+	kubeConfigFile      string
 
 	selfSignedCA    bool
 	selfSignedCAOrg string
@@ -85,8 +85,8 @@ func init() {
 	flags.StringVar(&opts.rootCertFile, "root-cert", "", "Specifies path to the root certificate file")
 
 	flags.StringVar(&opts.namespace, "namespace", "", "The namesapce that CA is running at.")
-	flags.BoolVar(&opts.targetAllNamespace, "all-namespace", false,
-		"Indicates whether to listen to all namespace or just the namespace CA is running at")
+	flags.BoolVar(&opts.targetAllNamespaces, "all-namespaces", false,
+		"Indicates whether to listen to all namespaces or just the namespace CA is running at")
 
 	flags.StringVar(&opts.kubeConfigFile, "kube-config", "",
 		"Specifies path to kubeconfig file. This must be specified when not running inside a Kubernetes pod.")
@@ -94,12 +94,10 @@ func init() {
 	flags.BoolVar(&opts.selfSignedCA, "self-signed-ca", false,
 		"Indicates whether to use auto-generated self-signed CA certificate. "+
 			"When set to true, the '--signing-cert' and '--signing-key' options are ignored.")
-
 	flags.StringVar(&opts.selfSignedCAOrg, "self-signed-ca-org", "k8s.cluster.local",
 		fmt.Sprintf("The issuer organization used in self-signed CA certificate (default to %s)",
 			selfSignedCAOrgDefault))
 
-	// TODO(wattli): change it to IntVar to make it configurable in yaml.
 	flags.DurationVar(&opts.caCertTTL, "ca-cert-ttl", defaultCACertTTL,
 		"The TTL of self-signed CA root certificate")
 	flags.DurationVar(&opts.certTTL, "cert-ttl", time.Hour, "The TTL of issued certificates")
@@ -137,7 +135,8 @@ func runCA() {
 	cs := createClientset()
 	ca := createCA(cs.CoreV1())
 	controllerNs := opts.namespace
-	if opts.targetAllNamespace {
+	if opts.targetAllNamespaces {
+		// Listen to all namespaces by specifying controllerNs to empty.
 		controllerNs = ""
 	}
 	sc := controller.NewSecretController(ca, cs.CoreV1(), controllerNs)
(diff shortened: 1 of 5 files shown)


pkg/pki/ca/ca.go
```diff
@@ -80,6 +105,47 @@ func NewSelfSignedIstioCA(caCertTTL, certTTL time.Duration, org string) (*IstioC
 // NewIstioCA returns a new IstioCA instance.
 func NewIstioCA(opts *IstioCAOptions) (*IstioCA, error) {
 	ca := &IstioCA{certTTL: opts.CertTTL}
+	// If the signing key/cert or root cert is empty, we should create a self-signed key/cert pair,
+	// and write it to the secret for persistent purpose.
+	// TODO(wattli): get rid of the NewSelfSignedIstioCA() after 0.2.
+	// K8s secret has a bug that you can't specify empty content, so we put a small content in it and
```
lookuptable: Can we link to the bug/issue here? 

Who is "we" when you say "we put a small content in it"?

pkg/pki/ca/ca.go
```diff
@@ -22,10 +22,29 @@ import (
...
+	IstioSecretType = "istio.io/key-and-cert"
+
+	// The ID/name for the certificate chain file.
+	CACertChainID = "ca-cert.pem"
+	// The ID/name for the private key file.
+	CAPrivateKeyID = "ca-key.pem"
+	// The ID/name for the CA root certificate file.
+	RootCertID = "root-cert.pem"
+
+	CAServiceAccount = "istio-ca-creds"
```
lookuptable: We can't assume the service account is a const, can we?

pkg/pki/ca/ca.go
```diff
@@ -80,6 +105,47 @@ func NewSelfSignedIstioCA(caCertTTL, certTTL time.Duration, org string) (*IstioC
...
+		// Rewrite the key/cert back to secret so they will be persistent when CA restarts.
+		secret := &apiv1.Secret{
+			Data: map[string][]byte{
+				CACertChainID:  pemCert,
+				CAPrivateKeyID: pemKey,
+				RootCertID:     pemCert,
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{serviceAccountNameAnnotationKey: CAServiceAccount},
+				Name:        SecretNamePrefix + CAServiceAccount,
```
lookuptable: So if an admin specifies an empty secret called `ca-signing-secret` and mounts it into the CA pod, the CA will in turn write the self-signed key/cert into a totally different secret (e.g. `istio-ca.istio-ca-creds`). This would be quite confusing, right? And it won't help in case the ca pod is recreated.

pkg/pki/ca/ca.go
```diff
@@ -22,10 +22,29 @@ import (
 	"errors"
 	"time"
 
+	"github.com/golang/glog"
 	"istio.io/auth/pkg/pki"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
lookuptable: I'd be careful in bringing dependency on K8S apis here. Previously CA is CA and all interactions with K8S are delegated to the `controller`.

pkg/pki/ca/ca.go
wattli: Good point, an easy way is to document it clearly not to mess up with the default CA config. It might worth to make it a flag going forward.

pkg/pki/ca/ca.go
wattli: It's actually not a service account, it's the secret name. Do we actually need this annotation?

pkg/pki/ca/ca.go
wattli: Good point. The secret (with dummy content) needs to be there before deploying istio-ca, and we will have a yaml file to create that secret (in front of istio-ca yaml). 

An easy way is to document it clearly not to mess up the default CA secret name (and I doubt people will actually do that). 

Of course this can be done via flag, but we will need admin to specify secret name at two places: where we do volume mount and CA input params. Personally I prefer not to introduce another flag to CA.

wattli:
@lookuptable  PTAL, thanks

cmd/istio_ca/main.go
```diff
@@ -172,14 +174,17 @@ func createCA() ca.CertificateAuthority {
 	caOpts := &ca.IstioCAOptions{
 		CertChainBytes:   certChainBytes,
 		CertTTL:          opts.certTTL,
+		Core:             core,
+		Namespace:        opts.namespace,
```
lookuptable: What happens when namespace is unspecified (i.e., in cluster-wide deployment)?

pkg/pki/ca/ca.go
lookuptable: The comment seems to be the same as before.

pkg/pki/ca/ca.go
lookuptable: hmm I feel it's quite a restriction to force the secret to be named as `isito-ca.istio-ca-creds`. 

IMO the idea workflow would be:

1. Admin creates a secret with an arbitrarily chosen name
1. The admin mounts the secret into CA at a well-defined dir
1. CA examines the dir, and write key/cert if the dir is empty
1. The content that CA writes is auto fed back to apiserver (we might need to verify if this is possible)

If step 4 is not possible, I'd vote to explicitly passing in a secret name.

pkg/pki/ca/ca.go
```diff
@@ -80,6 +107,47 @@ func NewSelfSignedIstioCA(caCertTTL, certTTL time.Duration, org string) (*IstioC
...
+		secret := &apiv1.Secret{
+			Data: map[string][]byte{
+				CACertChainID:  pemCert,
+				CAPrivateKeyID: pemKey,
+				RootCertID:     pemCert,
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{serviceAccountNameAnnotationKey: CAServiceAccount},
+				Name:        SecretNamePrefix + CAServiceAccount,
+				Namespace:   opts.Namespace,
```
lookuptable: This will like cause problem as Namespace could be empty.

wattli: Fixed.

cmd/istio_ca/main.go
wattli: Fixed.

wattli:
@lookuptable , please take a look, need to get this in ASAP. I will ask IBMer to review after your approval.

lookuptable:
I'll try to get to this tonight. Feel free to bring a second reviewer meanwhile to speed it up.

cmd/istio_ca/main.go
```diff
@@ -49,8 +50,9 @@ type cliOptions struct {
 	signingKeyFile  string
 	rootCertFile    string
 
-	namespace      string
-	kubeConfigFile string
+	namespace          string
+	targetAllNamespace bool
```
lookuptable: Nit: AllNamespaces

cmd/istio_ca/main.go
```diff
@@ -82,19 +84,22 @@ func init() {
...
 
 	flags.BoolVar(&opts.selfSignedCA, "self-signed-ca", false,
 		"Indicates whether to use auto-generated self-signed CA certificate. "+
 			"When set to true, the '--signing-cert' and '--signing-key' options are ignored.")
+
 	flags.StringVar(&opts.selfSignedCAOrg, "self-signed-ca-org", "k8s.cluster.local",
 		fmt.Sprintf("The issuer organization used in self-signed CA certificate (default to %s)",
 			selfSignedCAOrgDefault))
 
+	// TODO(wattli): change it to IntVar to make it configurable in yaml.
```
lookuptable: Sorry I don't agree with this todo. Right now you can already specify a string that can be parsed as a duration object (e.g. 5m, 10h)

cmd/istio_ca/main.go
```diff
@@ -121,13 +126,21 @@ func runCA() {
...
 
 	verifyCommandLineOptions()
 
-	ca := createCA()
 	cs := createClientset()
-	sc := controller.NewSecretController(ca, cs.CoreV1(), opts.namespace)
+	ca := createCA(cs.CoreV1())
+	controllerNs := opts.namespace
+	if opts.targetAllNamespace {
+		controllerNs = ""
```
lookuptable: Please add comment here

pkg/pki/ca/ca.go
```diff
@@ -22,15 +22,32 @@ import (
...
 
+	"github.com/golang/glog"
 	"istio.io/auth/pkg/pki"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
 )
 
 const (
+	// IstioSecretType is the Istio secret annotation type.
```
lookuptable: Not: all constants here can be unexported.

pkg/pki/ca/ca.go
```diff
@@ -22,15 +22,32 @@ import (
...
+
+	// CASecret stores the key/cert of self-signed CA for persistency purpose.
+	CASecret = "istio-ca-secret"
+
 	// The size of a private key for a self-signed Istio CA.
 	caKeySize = 2048
 )
 
-// CertificateAuthority contains methods to be supported by a CA.
+// CertificateAuthority contains methods to bfe supported by a CA.
```
lookuptable: Please revert this change.

pkg/pki/ca/ca.go
```diff
@@ -40,6 +57,7 @@ type CertificateAuthority interface {
 type IstioCAOptions struct {
 	CertChainBytes   []byte
 	CertTTL          time.Duration
+	Namespace        string
```
lookuptable: Guess you don't need this any more.

pkg/pki/ca/ca.go
```diff
@@ -56,24 +74,57 @@ type IstioCA struct {
...
 	opts := &IstioCAOptions{
-		CertTTL:          certTTL,
-		SigningCertBytes: pemCert,
-		SigningKeyBytes:  pemKey,
-		RootCertBytes:    pemCert,
+		CertTTL:   certTTL,
+		Namespace: namespace,
+	}
+	if err != nil {
+		glog.Warningf("Failed to get secret (error: %s), will create one", err)
```
lookuptable: Nit: this is a legit case so probably don't log it as warning?

pkg/pki/ca/ca.go
```diff
@@ -56,24 +74,57 @@ type IstioCA struct {
...
+				Namespace: namespace,
+			},
+			Type: IstioSecretType,
+		}
+		_, err := core.Secrets(opts.Namespace).Create(secret)
+		if err != nil {
+			glog.Errorf("Failed to create secret (error: %s)", err)
+			return nil, err
+		}
+	} else {
```
lookuptable: add a todo to verify secret content?

pkg/pki/ca/ca.go
```diff
@@ -56,24 +74,57 @@ type IstioCA struct {
...
+		opts.SigningCertBytes = pemCert
+		opts.SigningKeyBytes = pemKey
+		opts.RootCertBytes = pemCert
+
+		// Rewrite the key/cert back to secret so they will be persistent when CA restarts.
+		secret := &apiv1.Secret{
+			Data: map[string][]byte{
+				CACertChainID:  pemCert,
+				CAPrivateKeyID: pemKey,
+				RootCertID:     pemCert,
```
lookuptable: Nit: you don't need to save root cert into the secret as it's always identical to the ca cert.

pkg/pki/ca/ca.go
```diff
@@ -56,24 +74,57 @@ type IstioCA struct {
...
-		Org:          org,
-		IsCA:         true,
-		IsSelfSigned: true,
-		RSAKeySize:   caKeySize,
-	}
-	pemCert, pemKey := GenCert(options)
+func NewSelfSignedIstioCA(caCertTTL, certTTL time.Duration, org string, namespace string,
+	core corev1.CoreV1Interface) (*IstioCA, error) {
 
+	caSecret, err := core.Secrets(namespace).Get(CASecret, metav1.GetOptions{})
```
lookuptable: I feel it deserves some detailed comment about what is happening here.

cmd/istio_ca/main.go
wattli: done

cmd/istio_ca/main.go
wattli: I see, reverted.

cmd/istio_ca/main.go
wattli: done

pkg/pki/ca/ca.go
wattli: reverted.

pkg/pki/ca/ca.go
wattli: They are used in ca_test.go

lookuptable: That's in the sane package, right? Still don't need to export them.

lookuptable: Same

wattli: I see, just followed secret.go example, looks it used somewhere else.

pkg/pki/ca/ca.go
wattli: done

pkg/pki/ca/ca.go
wattli: Done

wattli:
@rshriram @ZackButcher , please take a look, this is not a breaking change. It would be great for this to make in 0.2 since customers do not need to manually generate some key/cert for our Istio CA any more. Otherwise it would be a bad user experience for customers.

cmd/istio_ca/main.go
```diff
@@ -82,12 +86,14 @@ func init() {
 	flags.StringVar(&opts.signingKeyFile, "signing-key", "", "Specifies path to the CA signing key file")
 	flags.StringVar(&opts.rootCertFile, "root-cert", "", "Specifies path to the root certificate file")
 
-	flags.StringVar(&opts.namespace, "namespace", "",
-		"Select a namespace for the CA to listen to. If unspecified, Istio CA tries to use the ${"+namespaceKey+"} "+
-			"environment variable. If neither is set, Istio CA listens to all namespaces.")
+	flags.StringVar(&opts.namespace, "namespace", "", "The namesapce that CA is running at.")
+
```
rshriram: Doesn't this break e2e ?

rshriram: Shouldn't this default to istio-system ?

cmd/istio_ca/main.go
```diff
@@ -121,13 +127,22 @@ func runCA() {
 		if value, exists := os.LookupEnv(namespaceKey); exists {
 			opts.namespace = value
 		}
+
+		if opts.namespace == "" {
+			glog.Fatal("Fail to get the namespace of this CA")
+		}
```
andraxylia: This will not work with the current templates in istio/istio, because for cluster-wide, namespace is passed as "".
It is too late to make this change in 0.2

cmd/istio_ca/main.go
wattli: Actually we don't have to, since it will check the env variable about the namespace (see line 127).

cmd/istio_ca/main.go
andraxylia: Suggestion is to make the namespace flags changes in 0.3, since it needs to cascade all the way to pilot and istio, and requires change in the yaml files.

The actual certificate changes lgtm.

wattli:
@costinm @andraxylia , please take a look, thanks.

cmd/istio_ca/main.go
```diff
@@ -85,6 +89,8 @@ func init() {
 	flags.StringVar(&opts.namespace, "namespace", "",
 		"Select a namespace for the CA to listen to. If unspecified, Istio CA tries to use the ${"+namespaceKey+"} "+
 			"environment variable. If neither is set, Istio CA listens to all namespaces.")
+	flags.StringVar(&opts.caNamespace, "ca-namespace", "istio-system", "Specify whether the current CA is running at.")
```
andraxylia: We fix one bug and we introduce UX limitation. With the current code and yaml templates, users can do a sed '/istio-system/kube-system' and still have a cluster wide installation. With this change, they will have to also start the CA with the flag --ca-namespace "kube-system". It is an acceptable trade-off as long as this gets documented. Even better would be to use POD_NAMESPACE for this.

Please replace "Specify whether the current CA is running at." with "Namespace where the Istio CA pods is running. If not specified, it defaults to POD_NAMESPACE".

In 0.3, we will need to swap the flags, since all controller use "namespace" mapped to POD_NAMESPACE, to indicate where the controller is running, and other flag names to indicate where they listen to. This will be a backward incompatible change, thus my recommendation was to do it properly in 0.2 , if we must fix the critical bug.

wattli: Thanks Andra for your input.

Comment of ca-namespace flag is updated.

Currently it's NAMESPACE, it's just one line change for me to use POD_NAMESPACE, do you think it is safe to change to POD_NAMESPACE for now?

Agreed on 0.3 vision.

cmd/istio_ca/main.go
wattli: It's reverted.

cmd/istio_ca/main.go
costinm: How about "Specify the namespace where Istio CA root CA is stored. Will not be used if explicit file or other storage mechanism is specified"

And rename it to 'istio-ca-storage-namespace' - to make it very clear it is specific to the CA storage (and avoid mixing it up with what Andra is concerned about )

pkg/pki/ca/ca.go
costinm: IMHO it's probably ok to directly use it - in future we can refactor when other storage mechanisms are introduced, but too early to abstract it.

pkg/pki/ca/ca.go
```diff
@@ -22,10 +22,24 @@ import (
...
+	"github.com/golang/glog"
 	"istio.io/auth/pkg/pki"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
 )
 
 const (
+	// istioSecretType is the Istio secret annotation type.
+	istioSecretType = "istio.io/key-and-cert"
```
costinm: 'key-and-cert- is too generic - maybe 'ca-root' ?

pkg/pki/ca/ca.go
```diff
@@ -56,24 +70,57 @@ type IstioCA struct {
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cASecret,
+				Namespace: namespace,
+			},
+			Type: istioSecretType,
+		}
+		_, err := core.Secrets(namespace).Create(secret)
+		if err != nil {
+			return nil, err
```
costinm: IMHO - log an error ("can't save the cert") - but keep running. Even if the CA restart, we will eventually recover - however if CA fails to start 
pretty much everything will stop working.

cmd/istio_ca/main.go
wattli: Comments and flag name are both updated

pkg/pki/ca/ca.go
wattli: Good point, done.

pkg/pki/ca/ca.go
wattli: done.

wattli:
Shriram approved it as well (somehow dismissed by one of my commit).

PR #268, opened 2017-09-20
Fix the systemd manifest to use the correct binary name

costinm:
```release-note
Fix startup for the debian files for standalone node agent.
```

### DIFF
--- a/tools/deb/istio-auth-node-agent.service
+++ b/tools/deb/istio-auth-node-agent.service
@@ -3,7 +3,7 @@ Description=istio-auth-node-agent: The Istio auth node agent
 Documentation=http://istio.io/
 
 [Service]
-ExecStart=/usr/local/bin/istio-auth-node-agent --logtostderr
+ExecStart=/usr/local/bin/node-agent --logtostderr
 Restart=always
 StartLimitInterval=0
 RestartSec=10


tools/deb/istio-auth-node-agent.service
```diff
@@ -3,7 +3,7 @@ Description=istio-auth-node-agent: The Istio auth node agent
 Documentation=http://istio.io/
 
 [Service]
-ExecStart=/usr/local/bin/istio-auth-node-agent --logtostderr
+ExecStart=/usr/local/bin/node-agent --logtostderr
```
ldemailly: that's a very generic name that shouldn't be as is in /usr/local/bin specially in a VM context

rshriram: Why is the name bing changed ?

ldemailly:
are you sure you have the right debian file ?
https://github.com/istio/auth/blob/master/bin/push-debian.sh
seems to use istio-auth-node-agent already

which test fails ? with which deb ?

I wouldn't push this as is

tools/deb/istio-auth-node-agent.service
ldemailly: apparently it's not trivial to rename in bazel (!) so in the meantime, I think it's cleaner anyway to have that file in an istio/ named directory : proposing #270

costinm:
It may be better if auth team would take over this - the problem is that the auth agent won't start, and the proper name and location are better decided by the auth team. This change is just changing the systemd config to use the binary name and location picked by auth. 

Since the  auth service is a spiffee agent and may not be istio specific or may need to be consistent with
spiffee - better for someone informed about all aspects (and future plans) to make the name or location change.

IMHO this is the safest fix for the immediate problem ( node not starting ), and code freeze is not the proper time to introduce new naming conventions/locations - but I'll trust the release managers .

ldemailly:
#270 is as safe and better (imo) for our users
but whichever

lookuptable:
/lgtm
/approve

#269 has been filed to make sure we change to a more proper binary name post V0.2 release.

PR #270, opened 2017-09-20
Fix systemd issue

ldemailly:
For now, without putting generic named binaries in user’s /usr/local/bin

Replacement for #268 

```release-note
Standalone node agent path fix (debian)
```


wattli:
Why not just use istio-auth-node-agent in the existing path?

tools/deb/BUILD
```diff
@@ -1,12 +1,13 @@
...
 
+// TODO: also rename the binary to istio-node-agent
 pkg_tar(
     name = "agent-bin",
     files = [
         "//cmd/node_agent",
     ],
     mode = "0755",
-    package_dir = "/usr/local/bin",
+    package_dir = "/usr/local/istio/bin",
```
costinm: -1 - consistency with envoy, pilot-agent. 
-1 on /usr/local/istio as well - /opt/istio is more consistent with other projects.

ldemailly:
> Why not just use istio-auth-node-agent in the existing path?

if you know how to do that, that'd be even better, I don't (kuat had some example but I wasn't sure how to plug it)

(though I still think we should leave /usr/local/bin of our users alone - except for binaries we expect them to run directly (like istioctl))

ldemailly:
A fix was urgently needed but not anymore?

tools/deb/BUILD
costinm: And I mean: envoy, pilot agents should be in the same istio path with auth agent, for consistency,
and if we want to move them from /usr/local/bin - a proper discussion in 0.3 frame would be more appropriate. It doesn't seem like renaming and moving binaries across the filesystem is best done in a 'code freeze'. 

Even 'envoy' is miss-leading, pilot-agent doesn't have istio- prefix if we want to consistently use this, etc.

ldemailly: envoy is envoy (though maybe we'd be smashing their pre existing envoy install so yes we should also not install ours in /usr/local/bin), pilot-agent is less likely to conflict than node-agent but I can make a PR to move pilot-agent to the same directory, but that's in pilot/ not auth/ repo so irrelevant here ?

because the other paths are broken isn't a reason to not fix this one, and we can fix the others easily too, that we didn't do it yet isn't a reason to use "consistency" as a way to block progress

costinm: Having istio binaries spread between /usr/local/bin and /usr/local/istio/bin (when the likely location in 0.3 will be /opt/istio - and we're still evaluating upgrade policies, i.e. if the directory needs to include a version to support smooth upgrades) is not a major progress.

And may not be the right thing to do while in freeze.

costinm: If Kuat or someone who know how to do the rename in few (safe) lines can rename the binary but leave it in /usr/local/bin - it would be great. Otherwise I would prefer #268. 

I don't like the node-agent name, but it's no worse than 'envoy' (which has the risk of conflicting with the real envoy) - and neither are critical release blockers.

ldemailly: one last time: we should move all our non user facing binaries out of /usr/local/bin and that's 1 line PR in each repo, we could get started here as we can't exactly make a PR across 3 repos to change it all at once

costinm: Sounds good for 0.3 - I'm not the release manager but I don't think it qualify as release blocker.

For the record - SPIFFE consistently uses "Node Agent" in their documentation, and if the goal of auth team is to provide the 'google implementation of spifee' - the name may be as intended, and location may be what the spiffe reference implementation uses. But this is the last thing we should be spending our time on 1 week after code freeze and before the release...

tools/deb/istio-auth-node-agent.service
```diff
@@ -3,7 +3,7 @@ Description=istio-auth-node-agent: The Istio auth node agent
 Documentation=http://istio.io/
 
 [Service]
-ExecStart=/usr/local/bin/istio-auth-node-agent --logtostderr
+ExecStart=/usr/local/istio/bin/node-agent --logtostderr
```
rshriram: Okay, this change is new. @kyessenov do we even have istio directory for pilot agents ? i thought we always had them in /usr/local/bin..

kyessenov: No, `/usr/local/blah` sounds odd to me. In terms of impact, these are just entry points in containers, so we can shuffle things if needed.

ldemailly: the point is in a VM we shouldn't put stuff with generic names (istioctl is fine, it's ours) in /usr/local/bin - this addresses that (and is unrelated to pilot-agent which can also be in the same directory in another PR but also doesn't have to be)

ldemailly: remember this is not containers folks, this is someone's actual pre-existing machine

tools/deb/BUILD
myidpt: We do not need to be aligned with SPIFFE in terms of implementation. :)
I have no strong preference on the binary location, but I think renaming is necessary. When do we plan to get this PR in? @ldemailly are you planning to rename the binary?

wattli:
@ldemailly , should we drop this?

ldemailly:
well we should have a unified location for istio stuff,
we could merge this for now/starter or drop it

up to you

ldemailly:
funny thing is I finally got around to try the full setup... and it doesn't work:

```
$ sudo systemctl status istio-auth-node-agent
● istio-auth-node-agent.service - istio-auth-node-agent: The Istio auth node agent
   Loaded: loaded (/lib/systemd/system/istio-auth-node-agent.service; disabled; vendor preset: ena
   Active: activating (auto-restart) (Result: exit-code) since Fri 2017-10-13 20:20:45 UTC; 9s ago
     Docs: http://istio.io/
  Process: 3774 ExecStart=/usr/local/bin/node-agent --logtostderr (code=exited, status=203/EXEC)
 Main PID: 3774 (code=exited, status=203/EXEC)
```
the binaries are : 
```
ldemailly@demo-vm-1:~$ ls -l /usr/local/bin/
envoy              istio-iptables.sh  istio-start.sh     node_agent         pilot-agent
```

tools/deb/istio-auth-node-agent.service
```diff
@@ -3,7 +3,7 @@ Description=istio-auth-node-agent: The Istio auth node agent
 Documentation=http://istio.io/
 
 [Service]
-ExecStart=/usr/local/bin/node-agent --logtostderr
+ExecStart=/usr/local/istio/bin/node_agent --logtostderr
```
ldemailly: important part here is the `_` and not `-`

ldemailly:
tested with export AUTH_DEBIAN_URL="https://storage.googleapis.com/istio-artifacts/auth/0c7d945b0365d8ce4adea1cbc6c6ae9d30d1bfd7/artifacts/debs"

PR #281, opened 2017-10-05
Add preliminary AWS support for node agent

crazytan:
First step in supporting AWS. To make it truly work we'll need some refactoring in the codebase. For example, `na.createRequest()` needs to be pluggable to correctly populate `NodeAgentCredentials` field in request message.
```release-note
none
```

### DIFF
--- a/cmd/node_agent/na/aws.go
+++ b/cmd/node_agent/na/aws.go
@@ -48,7 +48,7 @@ func (na *awsPlatformImpl) IsProperPlatform() bool {
 func (na *awsPlatformImpl) GetServiceIdentity() (string, error) {
 	userdata, err := na.fetcher.GetUserData()
 	if err != nil {
-		return "", fmt.Errorf("Failed to get service identity: %v", err)
+		return "", fmt.Errorf("Failed to get EC2 user data: %v", err)
 	}
 	var dat map[string]string
 	err = json.Unmarshal([]byte(userdata), &dat)
--- a/cmd/node_agent/na/aws.go
+++ b/cmd/node_agent/na/aws.go
@@ -1,0 +1,59 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package na
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	cred "istio.io/auth/pkg/credential"
+)
+
+type awsPlatformImpl struct {
+	fetcher *cred.AwsTokenFetcher
+}
+
+// This is extracted from the same function in gcp.go.
+// Should put the common logic elsewhere in the future.
+func (na *awsPlatformImpl) GetDialOptions(cfg *Config) ([]grpc.DialOption, error) {
+	creds, err := credentials.NewClientTLSFromFile(cfg.RootCACertFile, "")
+	if err != nil {
+		return nil, err
+	}
+
+	options := []grpc.DialOption{grpc.WithTransportCredentials(creds)}
+	return options, nil
+}
+
+func (na *awsPlatformImpl) IsProperPlatform() bool {
+	return na.fetcher.IsOnAWS()
+}
+
+// Extract service identity from userdata. This function should be
+// pluggable for different AWS deployments in the future.
+func (na *awsPlatformImpl) GetServiceIdentity() (string, error) {
+	userdata, err := na.fetcher.GetUserData()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get service identity: %v", err)
+	}
+	var dat map[string]string
+	err = json.Unmarshal([]byte(userdata), &dat)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


crazytan:
I signed it!

pkg/credential/token.go
```diff
@@ -30,12 +35,49 @@ type GcpTokenFetcher struct {
 	Aud string
 }
 
+// AwsTokenFetcher implements the token fetcher in AWS.
+// It gets the EC2 instance identity document as described
+// in the doc: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+type AwsTokenFetcher struct {
```
myidpt: Can we rename all TokenFetchers to IdentityDocFethers, to be more precise?

pkg/credential/token.go
```diff
@@ -30,12 +35,49 @@ type GcpTokenFetcher struct {
...
+}
+
+// IsOnAWS returns if node agent is running on AWS EC2 instance
+func (fetcher *AwsTokenFetcher) IsOnAWS() bool {
```
myidpt: Make IsOnAWS private?

cmd/node_agent/na/aws.go
```diff
@@ -0,0 +1,59 @@
...
+	cred "istio.io/auth/pkg/credential"
+)
+
+type awsPlatformImpl struct {
+	fetcher *cred.AwsTokenFetcher
+}
+
+// This is extracted from the same function in gcp.go.
+// Should put the common logic elsewhere in the future.
+func (na *awsPlatformImpl) GetDialOptions(cfg *Config) ([]grpc.DialOption, error) {
```
myidpt: I think you may need to pass the identity document to CA so that the CA can authenticate the node agent.

cmd/node_agent/na/aws.go
```diff
@@ -0,0 +1,59 @@
...
+// pluggable for different AWS deployments in the future.
+func (na *awsPlatformImpl) GetServiceIdentity() (string, error) {
+	userdata, err := na.fetcher.GetUserData()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get service identity: %v", err)
+	}
+	var dat map[string]string
+	err = json.Unmarshal([]byte(userdata), &dat)
+	if err != nil {
+		return "", fmt.Errorf("Failed to get service identity: %v", err)
```
wattli: Can we use 'Failed to unmarshal service identity' to differentiate from the log above. This will make the debugging easier.

cmd/node_agent/na/aws.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	return dat["APIGEE_POD"], nil
```
wattli: Make APIGEE_POD a const? And where is it used?

pkg/credential/token.go
wattli: I think that can go with a follow-up PR. That will make this PR clean and easier to track :)

pkg/credential/token.go
```diff
@@ -30,12 +35,49 @@ type GcpTokenFetcher struct {
...
+// be fixed in upstream.
+func (fetcher *AwsTokenFetcher) FetchToken() (string, error) {
+	sess := session.Must(session.NewSession())
+	svc := ec2metadata.New(sess)
+	if svc.Available() {
+		doc, err := svc.GetInstanceIdentityDocument()
+		if err == nil {
+			bytes, _ := json.Marshal(doc)
+			return string(bytes), nil
+		}
```
wattli: do you want to have a else block to handle err != nil case?

pkg/credential/token.go
myidpt: Yes, that works :)

cmd/node_agent/na/aws.go
crazytan: True. But as I mentioned above, right now my understanding is that identity document should be passed in the `NodeAgentCredentials` field. However current implementation doesn't support overriding `na.createRequest()`. I think I'll rebase this PR once refactoring is done.

cmd/node_agent/na/aws.go
crazytan: agree.

pkg/credential/token.go
crazytan: if `err != nil` it will fall back to end of the function which throws an error.

cmd/node_agent/na/aws.go
crazytan: Just here. My concern is that even for AWS we may want pluggable interface for users to override the logic here, since on AWS there's no service account, so there is no standard way of extracting the service identity. I too feel uncomfortable to use `APIGEE_POD` here.

cmd/node_agent/na/aws.go
myidpt: Sounds good to me.

cmd/node_agent/na/aws.go
crazytan: The detailed error is logged with `%v", err`

pkg/credential/token.go
crazytan: this is called by `awsPlatformImpl.IsProperPlatform()`. Will move it to better place.

pkg/credential/token.go
wattli: But you will lose the debugging info in 'err', right?

cmd/node_agent/na/aws.go
wattli: I mean this will make the debugging easier since 1) generally speaking, the detailed error of two functions can give you the same error message, you don't want to count on others :-)  2) even the detailed errors are different, you need to read into the them and guess which function cause the error.

No big deal, i will let you decide.

pkg/credential/token.go
crazytan: good catch! I'll change it

cmd/node_agent/na/aws.go
crazytan: Makes sense! I'll change it in case something funny happens in aws call...

PR #296, opened 2017-10-10
Fix integration test failure due to RBAC enablement

myidpt:
**Release note**:

```release-note
none
```

### DIFF
--- a/integration/main.go
+++ b/integration/main.go
@@ -160,19 +160,6 @@ func deleteTestNamespace(clientset kubernetes.Interface) {
 }
 
 func deployIstioCA(clientset kubernetes.Interface) {
-	// Create ServiceAccount
-	serviceaccount := v1.ServiceAccount{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "v1",
-		},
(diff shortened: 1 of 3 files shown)


integration/main.go
```diff
@@ -152,13 +153,64 @@ func createTestNamespace(clientset kubernetes.Interface) string {
...
+			APIVersion: "rbac.authorization.k8s.io/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "istio-ca-role",
+			Namespace: opts.namespace,
+		},
+		Rules: []v1alpha1.PolicyRule{
+			{
+				Verbs:     []string{"create", "get", "watch", "list", "update"},
+				APIGroups: []string{"core", ""},
```
lookuptable: Do we need the empty APIGroup (`""`)?

myidpt: Actually empty APIGroups means "core" when applied via yaml file, but the K8s API does not allow calling with empty field. Also, specifying "core" alone will cause "unknown user" error. I think this may be a bug in K8s. This is the working solution for now.

PR #300, opened 2017-10-11
Add Makefile and use fixed linter

sebastienvas:
```release-note
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 SHELL := /bin/bash
 
-setup:
+setup: build
 	@bin/setup.sh
 
 build:
@@ -26,11 +26,11 @@ clean:
 test:
 	@bazel test //...
 
-lint: build
+lint: setup
 	@bin/linters.sh
 	@bin/headers.sh
 
-coverage:
+coverage: setup
 	@bin/codecov.sh
 
 racetest:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,39 @@
+## Copyright 2016 Istio Authors
+##
+## Licensed under the Apache License, Version 2.0 (the "License");
+## you may not use this file except in compliance with the License.
+## You may obtain a copy of the License at
+##
+##     http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+test:
+	@bazel test //...
+
+lint: build
+	@bin/linters.sh
+	@bin/headers.sh
+
+coverage:
+	@bin/codecov.sh
```
lookuptable: coverage depends on setup I believe

Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+clean:
+	@bazel clean
+
+test:
+	@bazel test //...
+
+lint: build
+	@bin/linters.sh
+	@bin/headers.sh
```
lookuptable: It might make more sense to create a separate target for headers but this is really a nit

sebastienvas: I am just following mixer's one. Such that all makefile behave the same.

Makefile
sebastienvas: Done

bin/linters.sh
```diff
@@ -1,8 +1,15 @@
...
-gometalinter --concurrency=4 --enable-gc --deadline=300s --disable-all\
-  --enable=maligned\
+LINTER_SHA='bfcc1d6942136fd86eb6f1a6fb328de8398fbd80'
+
+docker run \
+  -v $(bazel info output_base):$(bazel info output_base) \
+  -v $(pwd):/go/src/istio.io/auth \
+  -w /go/src/istio.io/auth gcr.io/istio-testing/linter:${LINTER_SHA} \
+  --concurrency=4 --enable-gc --deadline=300s --disable-all\
+  --enable=aligncheck\
```
myidpt: I think aligncheck is deprecated in favor of maligned:
https://github.com/alecthomas/gometalinter/pull/367

prow/auth-presubmit.sh
```diff
@@ -40,7 +40,6 @@ else
 fi
 
 echo "=== Bazel Build ==="
-./bin/install-prereqs.sh
```
myidpt: Is it always true that we don't need it?

bin/linters.sh
sebastienvas: Yes but the version that we use in the docker image is older than this.

prow/auth-presubmit.sh
sebastienvas: this just install gometalinter which is not needed.

PR #301, opened 2017-10-14
Fix auth node agent start error issue (in release-0.2 branch)

ldemailly:
Cherry pick of #270 

Fix auth node agent start error issue (wrong file name)

Fixes `_` /`-` typo in binary name

also without putting generic named binaries in user’s /usr/local/bin

Actual fix for #268 

```release-note
Standalone node agent path fix (debian)
```

(cherry picked from commit f6efb0ab402ae55b7ba5b7b249e3df312422258d)

PR #304, opened 2017-10-14
Fix integration test failure due to RBAC enablement (#296)

wattli:
Automatic merge from submit-queue

Fix integration test failure due to RBAC enablement

**Release note**:

```release-note
none
```

**Release note**:

```release-note
```

### DIFF
--- a/bin/linters.sh
+++ b/bin/linters.sh
@@ -2,7 +2,7 @@
 set -ex
 
 gometalinter --concurrency=4 --enable-gc --deadline=300s --disable-all\
-  --enable=aligncheck\
+  --enable=maligned\
   --enable=deadcode\
   --enable=errcheck\
   --enable=gas\
--- a/integration/main.go
+++ b/integration/main.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/rbac/v1alpha1"
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/golang/glog"
@@ -102,7 +103,7 @@ func runTests(cmd *cobra.Command, args []string) {
 
 	// Delete the secret.
 	do := &metav1.DeleteOptions{}
-	if err := opts.clientset.Core().Secrets(opts.namespace).Delete("istio.default", do); err != nil {
+	if err := opts.clientset.CoreV1().Secrets(opts.namespace).Delete("istio.default", do); err != nil {
 		glog.Fatal(err)
 	} else {
 		glog.Info(`Secret "istio.default" has been deleted`)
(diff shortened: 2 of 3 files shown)


integration/main.go
```diff
@@ -152,13 +153,64 @@ func createTestNamespace(clientset kubernetes.Interface) string {
...
 	name := opts.namespace
-	if err := clientset.Core().Namespaces().Delete(name, &metav1.DeleteOptions{}); err != nil {
+	if err := clientset.CoreV1().Namespaces().Delete(name, &metav1.DeleteOptions{}); err != nil {
 		glog.Fatalf("failed to delete namespace %q (error: %v)", name, err)
 	}
 	glog.Infof("Namespace %v is deleted", name)
 }
 
 func deployIstioCA(clientset kubernetes.Interface) {
+	// Create Role
```
ldemailly: why isn’t this a yaml like the other ? (why create in code vs yaml)

ldemailly:
but I guess that's however it was in the trunk too

I still don't understand how we made 0.2.7 without needing this ? but...

/lgtm

bin/linters.sh
```diff
@@ -2,7 +2,7 @@
 set -ex
 
 gometalinter --concurrency=4 --enable-gc --deadline=300s --disable-all\
-  --enable=aligncheck\
+  --enable=maligned\
```
ldemailly: this part will conflict with seb's pr I think

ldemailly:
so I guess we merge this one and then #301 will pass 

can we maybe just add a commit in #301 directly?

@ayj I'll let you decide

bin/linters.sh
myidpt: seb's PR is based on an old version of gometalinter instead of the HEAD. aligncheck has changed to maligned recently.
I think we should keep this change unless we cherrypick seb's PR as well.

integration/main.go
myidpt: To me that's a minor question. We can change the code to yaml files easily in the future. The tests are the same. :)

myidpt:
@ldemailly When was 0.2.7 made? The testing cluster switched to enabling RBAC last week.

ldemailly:
oh that explains ! didn't know auth had a different non rbac cluster

bin/linters.sh
ldemailly: I cherry picked seb already in my PR #301 ... but maybe I can un cherry pick or resolve conflicts

integration/main.go
ldemailly: sure

bin/linters.sh
myidpt: I see. I think we may still need this to pass the tests for this PR.

ldemailly:
any objection to getting this PR into the release branch asap so I can rebase my modified #301 and we get that one in too ? @ayj

bin/linters.sh
ldemailly: it's ok, I dropped seb's change from my PR; this one is the correct minimal one and it won't conflict anymore

ldemailly:
merging this as it's needed for anything at all to pass in that branch with the test env change

PR #309, opened 2017-10-20
add signature verification and unit tests for AWS plugin

crazytan:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/node_agent/na/aws.go
+++ b/cmd/node_agent/na/aws.go
@@ -15,12 +15,39 @@
 package na
 
 import (
+	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/fullsailor/pkcs7"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"istio.io/auth/pkg/pki"
+)
+
+const (
+	// AWSCertificatePem is the official public certificate for AWS
+	// copied from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+	AWSCertificatePem = `-----BEGIN CERTIFICATE-----
+MIIC7TCCAq0CCQCWukjZ5V4aZzAJBgcqhkjOOAQDMFwxCzAJBgNVBAYTAlVTMRkw
+FwYDVQQIExBXYXNoaW5ndG9uIFN0YXRlMRAwDgYDVQQHEwdTZWF0dGxlMSAwHgYD
+VQQKExdBbWF6b24gV2ViIFNlcnZpY2VzIExMQzAeFw0xMjAxMDUxMjU2MTJaFw0z
+ODAxMDUxMjU2MTJaMFwxCzAJBgNVBAYTAlVTMRkwFwYDVQQIExBXYXNoaW5ndG9u
+IFN0YXRlMRAwDgYDVQQHEwdTZWF0dGxlMSAwHgYDVQQKExdBbWF6b24gV2ViIFNl
+cnZpY2VzIExMQzCCAbcwggEsBgcqhkjOOAQBMIIBHwKBgQCjkvcS2bb1VQ4yt/5e
+ih5OO6kK/n1Lzllr7D8ZwtQP8fOEpp5E2ng+D6Ud1Z1gYipr58Kj3nssSNpI6bX3
+VyIQzK7wLclnd/YozqNNmgIyZecN7EglK9ITHJLP+x8FtUpt3QbyYXJdmVMegN6P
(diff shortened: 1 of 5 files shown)


cmd/node_agent/na/aws.go
```diff
@@ -41,16 +68,44 @@ func (na *awsPlatformImpl) IsProperPlatform() bool {
...
-// pluggable for different AWS deployments in the future.
+// This function should be pluggable for different AWS deployments.
 func (na *awsPlatformImpl) GetServiceIdentity() (string, error) {
 	return "", nil
 }
 
+func (na *awsPlatformImpl) GetInstanceIdentityDocument() ([]byte, error) {
+	cert, err := pki.ParsePemEncodedCertificate([]byte(AWSCertificatePem))
+	if err != nil {
+		return []byte{}, fmt.Errorf("Failed to parse AWS public certificate: %v", err)
```
wattli: what about return nil instead of []byte{} ?

wattli: same below

cmd/node_agent/na/aws.go
```diff
@@ -15,12 +15,39 @@
...
+	"github.com/fullsailor/pkcs7"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"istio.io/auth/pkg/pki"
+)
+
+const (
+	// AWSCertificatePem is the official public certificate for AWS
+	// copied from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+	AWSCertificatePem = `-----BEGIN CERTIFICATE-----
```
myidpt: Could you put this content into a file like cert_aws.pem in testdata/? Thanks.

cmd/node_agent/na/aws.go
```diff
@@ -41,16 +68,44 @@ func (na *awsPlatformImpl) IsProperPlatform() bool {
...
 }
 
-// Extract service identity from userdata. This function should be
-// pluggable for different AWS deployments in the future.
+// This function should be pluggable for different AWS deployments.
 func (na *awsPlatformImpl) GetServiceIdentity() (string, error) {
 	return "", nil
 }
 
+func (na *awsPlatformImpl) GetInstanceIdentityDocument() ([]byte, error) {
```
myidpt: Do you want to make this function private?

cmd/node_agent/na/aws_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	server.Close()
+	if na.IsProperPlatform() {
+		t.Errorf("On Proper Platform: expected false")
+	}
+}
+
+func TestGetInstanceIdentityDocument(t *testing.T) {
+	server := initTestServer(
+		"/latest/dynamic/instance-identity/pkcs7",
```
myidpt: If possible, can you move this to a file in testdata/?

cmd/node_agent/na/aws.go
crazytan: Done.

cmd/node_agent/na/aws.go
crazytan: but it's not just used for test; it'll be used in production env to verify signature

cmd/node_agent/na/aws_test.go
crazytan: will do

cmd/node_agent/na/aws.go
crazytan: will do

cmd/node_agent/na/aws.go
myidpt: You are right. Maybe put it in a file in a new dir data/?

crazytan: But wouldn't it require the data folder to be deployed alongside the Go binary for deployment?

myidpt: Yes, but will that be a problem?

myidpt: Anyway, we can change this later.

crazytan: My concern is that it kind of goes against the deployment model of Go (take a binary and deploy it any where). If putting the cert in the code is not desired, we can add a cmd option for this.
The deep issue is that AWS doesn't provide a way for us to dynamically retrieve the public cert, nor does it state whether the cert would be rotated in the future.

myidpt: Thanks for the explanation.

cmd/node_agent/na/aws.go
crazytan: I'll change it in next PR

cmd/node_agent/na/aws_test.go
crazytan: I'll change it in next PR

PR #310, opened 2017-10-23
Refactor node agent.

myidpt:
**Release note**:

```release-note
Move platform related content from cmd/node_agent/na to pkg/platform
```


cmd/node_agent/main.go
```diff
@@ -41,15 +41,15 @@ func init() {
...
-	flags.StringVar(&naConfig.CertChainFile, "cert-chain",
-		"/etc/certs/cert-chain.pem", "Node Agent identity cert file")
-	flags.StringVar(&naConfig.KeyFile,
-		"key", "/etc/certs/key.pem", "Node identity private key file")
 	flags.StringVar(&naConfig.IstioCAAddress,
 		"ca-address", "istio-ca:8060", "Istio CA address")
-	flags.StringVar(&naConfig.RootCACertFile, "root-cert",
-		"/etc/certs/root-cert.pem", "Root Certificate file")
 	flags.StringVar(&naConfig.Env, "env", "onprem", "Node Environment : onprem | gcp | aws")
+	flags.StringVar(&naConfig.PlatformConfig.CertChainFile, "cert-chain",
```
crazytan: just a thought: is it possible to enable a custom flagset for different platform? Because for other environments, the node agent bootstrap tokens may not be cert/key pair.

crazytan: just a thought: is it possible to enable a custom flagset for different platform? Because for other environments, the node agent bootstrap tokens may not be cert/key pair.

pkg/platform/client.go
```diff
@@ -0,0 +1,63 @@
...
+package platform
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+
+	"google.golang.org/grpc"
+	cred "istio.io/auth/pkg/credential"
```
crazytan: why not merge this with `platform/gcp`

cmd/node_agent/na/nafactory.go
```diff
@@ -41,22 +39,17 @@ func NewNodeAgent(cfg *Config) (NodeAgent, error) {
...
+	} else {
+		return nil, err
 	}
 
 	cAClient := &cAGrpcClientImpl{}
 	na.cAClient = cAClient
 
 	secretServer, err := workload.NewSecretServer(
-		workload.NewSecretFileServerConfig(cfg.CertChainFile, cfg.KeyFile))
+		workload.NewSecretFileServerConfig(cfg.PlatformConfig.CertChainFile, cfg.PlatformConfig.KeyFile))
```
crazytan: Is it right to pass `CertChainFile` and `KeyFile` here? I thought these options are for node agent identity.

cmd/node_agent/main.go
myidpt: I'm not sure. I think that can be a TODO.

cmd/node_agent/na/nafactory.go
myidpt: I also think this is problematic. It only works on single-tenant on-prem VMs, where the node agent identity is the same as the service identity. But for other cases, this should be different. Will add a TODO here.

cmd/node_agent/na/BUILD
```diff
@@ -4,29 +4,21 @@ go_library(
...
         "util.go",
     ],
     visibility = ["//visibility:public"],
     deps = [
-        "//pkg/credential:go_default_library",
         "//pkg/pki:go_default_library",
         "//pkg/pki/ca:go_default_library",
-        "//pkg/util:go_default_library",
+        "//pkg/platform:go_default_library",
+        "//pkg/platform/mock:go_default_library",
```
crazytan: Does it need mock package? This will build mock client into the node agent binary.

myidpt: No, it should be in the test package. Thanks for catching this.

cmd/node_agent/main.go
wattli: Good suggestion, it can be a TODO.

Issue #294, opened 2017-10-10
Define plugin interface for node agent 


crazytan:
There are several problems in current node agent implementation:

1. `type NodeAgent interface { Start() error }` is only implemented by `nodeAgentInternal` and there's no way in the code to choose different implementations.
2. instead, cmd argument `--env` chooses which `platformSpecificRequest` implementation to use in as `nodeAgentInternal.platformSpecificRequest` field. However, it doesn't provide custom `createRequest()` function.
3. even for the same platform, different users may have different logic, e.g., for `GetServiceIdentity()`.

PR #314, opened 2017-10-27
decouple node agent config from platform config (#294)

crazytan:
**Release note**:

```release-note
NONE
```
(Sorry this PR is a little large...) 

1. Cmd flag parsing now happens in two phases: first flags for node agent are parsed; then platform flags are added based on the env flag and `os.Args` are parsed again.
2. `PlatformConfig` is now an interface with `GetFlagSet()` method. Its implementation should also contain platform config variables which are automatically populated by `pflag` library. (See `GetFlagSet()` implementation.)
3. To make compilation work, I also added `--service-cert` and `--service-priv-key` options for node agent.

A possible break change is that platform flags have to be specified after node agent flags, otherwise the node agent flag parsing would fail. This is due to the limit in `pflag` library that it can't skip unrecognized flags.

### DIFF
--- a/cmd/node_agent/main.go
+++ b/cmd/node_agent/main.go
@@ -20,8 +20,8 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
-
 	flag "github.com/spf13/pflag"
+
 	"istio.io/auth/cmd/node_agent/na"
 	"istio.io/auth/pkg/cmd"
 	"istio.io/auth/pkg/platform"
--- a/cmd/node_agent/main.go
+++ b/cmd/node_agent/main.go
@@ -15,13 +15,16 @@
 package main
 
 import (
+	"io/ioutil"
 	"os"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 
+	flag "github.com/spf13/pflag"
 	"istio.io/auth/cmd/node_agent/na"
 	"istio.io/auth/pkg/cmd"
+	"istio.io/auth/pkg/platform"
 )
 
 var (
@@ -29,7 +32,7 @@ var (
 
 	rootCmd = &cobra.Command{
 		Run: func(cmd *cobra.Command, args []string) {
-			runNodeAgent()
+			runNodeAgent(cmd.Flags())
 		},
 	}
 )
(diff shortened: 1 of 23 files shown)


cmd/node_agent/main.go
```diff
@@ -61,7 +61,23 @@ func main() {
...
+	cc, err := platform.NewClientConfig(naConfig.Env)
+	if err != nil {
+		glog.Error(err)
+		os.Exit(-1)
+	}
+
+	// parse custom flags
+	cFlags := cc.GetFlagSet()
+	cFlags.AddFlagSet(flags)
+	err = cFlags.Parse(os.Args[1:])
```
myidpt: Please make sure you have tested this manually. Also, can you write unit test for this function?

pkg/platform/client.go
```diff
@@ -44,15 +40,29 @@ type Client interface {
...
 // NewClient is the function to create implementations of the platform metadata client.
-func NewClient(platform string, config ClientConfig, caAddr string) (Client, error) {
+func NewClient(platform string, cfg ClientConfig, caAddr string) (Client, error) {
 	switch platform {
 	case "onprem":
-		return NewOnPremClientImpl(config.CertChainFile), nil
+		return NewOnPremClientImpl(cfg), nil
 	case "gcp":
-		return NewGcpClientImpl(caAddr), nil
+		return NewGcpClientImpl(cfg, caAddr), nil
```
myidpt: Could you merge caAddr into GCP client cfg?

pkg/platform/client.go
```diff
@@ -17,22 +17,18 @@ package platform
 import (
 	"fmt"
 
+	flag "github.com/spf13/pflag"
```
myidpt: It seems strange to put flag into this platform package. Can you try to move the flags to pkg/cmd/?

Also, as you mentioned in the PR description:
"A possible break change is that platform flags have to be specified after node agent flags, otherwise the node agent flag parsing would fail."
I'm afraid this will cause trouble for our customers :( Can you please try to make the flags independent of orders? Otherwise I'm afraid we have to stick to the original approach until we find some better solution.

myidpt (changes requested):

pkg/platform/client.go
crazytan: Hmmm I'll try to see if I can work around this issue.

crazytan: I'll submit a separate PR for flag processing and testing, and then rebase this PR on top of that. It will allow node agent flags and platform flags mixed in order.

myidpt: Thank you for the work! :)

myidpt:
This repo is deprecated. Please create a PR in istio/istio instead.
Istio Security is moved to: https://github.com/istio/istio/tree/master/security

crazytan:
@myidpt Yes I'm aware of this as I was waiting for the code merge to be completed this week.

geeknoid:
This repo is no longer accepting PRs. Please resubmit this change to the istio/istio repo.

Thanks.

